PR #1, opened 2016-03-06
First cut for the Bloomberg Kafka connect source

stheppi:
Adding the Bloomberg Source for Kafka Connect. Currently the BEmu.jar doesn't work with the latest API.


andrewstevenson:
Great work Stefan! Upgrading the Emulator would be brilliant.

PR #2, opened 2016-03-09
Bloomberg Emulator  + Payload Schema

stheppi:
Added the jar and source code for the Bloomberg emulator. To use the emulator is a case of changing the dependency in the pom (Is commented out the emulator)

Added support to generate the avro schema and connect schema from Bloomberg Data. 
 I need to add configuration options to be able to switch from plain json to jsonconverter when converting between BloombergData and SourceRecord.


andrewstevenson:
I think we should wrap the avro fields in a union of [null, type] and set the default to allow schema evolution? 

Converting the json to avro with unions is a little tricky as the jsonencoder needs the type in the json string like {"type": "string", "symbol" : "APPL"}

andrewstevenson:
Also is the jsonConverter provided by Kafka of no use? [https://github.com/apache/kafka/tree/trunk/connect/json/src/main/java/org/apache/kafka/connect/json](https://github.com/apache/kafka/tree/trunk/connect/json/src/main/java/org/apache/kafka/connect/json)

andrewstevenson:
Build fail?

[https://travis-ci.org/andrewstevenson/stream-reactor/builds/114668564](https://travis-ci.org/andrewstevenson/stream-reactor/builds/114668564)

stheppi:
strange. looks like I can't run tests for one of the modules

mvn test -pl $module  will trigger the error.

Anyway I fixed the unit test. it was a mock related issue where I wasn't
mocking a method.


I am a little bit confused on  the payload side. Avro4s - json doesn't
convert the json to avro but infers the schema. I created the support to
build the schema without first going to json (it won't be efficient).
Second part is  we have taken the approach each ticker can have their own
fields. Therefore each message is self contained: msg1 has field1,field2
and msg2 has field3,field4. Adding support for schema generation in this
case I can't see a reason. Maybe we should specify a schema for all the
fields possible-> although with the nested values I have no clue what to
add.

Furthermore I have a map of String-> Object. Looking at the avro serializer
it doesn't seem it will work. I will try it.  The JsonConverter seems to be
the option. I can build the connect.schema and pass the json to the
converter to create the byte (yes not ideal on size) but at least this way
it has the schema going along. A schema that one could infer anyway since
it is based on the json fields available.


2016-03-09 8:14 GMT+00:00 Andrew Stevenson <notifications@github.com>:

andrewstevenson:
Avro4s infers a schems. Kite SDK does the same and also converts to avro from json.

What if create on uber schema for the fields we support based on that constant class?

It probably makes sense to partition per ticker therefore i believe we can support evolving schemas per key.

I'm not convinced by Kafka's converters, there's some chatter they aren't correctly.
________________________________

stheppi:
This support is done. However what I don't like is the fact it has to go
through json and then from json using the schema to avro.

Going back to the schema and tickers.Creating a uber schema is an option.
However, as I think I said previously there are fields which yield complex
types presumably (for data type Sequence or CHoice if you look at the
BloombergFieldValueFn it is recursive. so then I have no clue what to add.
we can say we limit to just those fields in the constants. What do you say?
Currently the schema I build is from the data I receive. There is no need
for union with null since there is no null - I exclude those fields from
Bloomberg if they have null value. Partitioning by key is an option but you
can have tiker1 and ticker2 hashing to the same partition but they have
different fields. So you can't assume the schema is the same for all the
messages on that partition.

THe first commit was using json on the basis that someone for example just
inserts into Cassandra via the sink :). so not much processing to be done.

So I will modify the source now to allow json payload or avro with each
message having their own schema. if you think having the uber schema is
better I will happily code that but then we ignore Bloomberg fields that
are Sequence/Choice that will provide nested types.

Let me know what you think


2016-03-09 17:22 GMT+00:00 Andrew Stevenson <notifications@github.com>:

stheppi:
Cassandra sink tests fail but doesn't say which one

andrewstevenson:
Try randomly do on Travis. Need to figure out why. I'll rerun.
________________________________

andrewstevenson:
I suppose we need to think about the destination sink. If Cassandra it will take the struct Connect schema and concert to json. If this struct contains only one field which is a string then it expects a table in Cassandra with one field.

What happens for hdfs? We should make sure we can evolve the schemas hence the union fields.

Maybe I missed something but I thought Bloomberg spat out strings. If not we need a converter like the jdbc source. I agree to jdin to avro feels wrong.
________________________________

stheppi:
I think I can build a genericrecord out of the map and that can be
serialized to avro. therefore we avoid the json step.

In the documentation from blmbg they specify each field type. it can be
string, float, double, int, datetime(their implementation) or time (their
implementation).

I can easily create the uber schema. and work with that where all fields
are optional. should we go with that?

If you read the readme changes I have added the ticker to the content. but
I think it shouldn't be
{
"ticker": value
"fields": {
  "field1" : "value1",
  "field2": "value2",
  ...
}

but rather
{
  "ticker": "value",
  "field1": "value",
  "filed2": "value",
  ...
}

It will fit better with storing to Cassandra for example. What do you think?


2016-03-10 6:28 GMT+00:00 Andrew Stevenson <notifications@github.com>:

andrewstevenson:
I’m tempted by the uber schema but it could become hard to manage. If we can convert the BB map to a flat schema like your latter example I think this is cleaner so lets’ not go uber until we require it.

All fields should be nullable (null first in union) with a default of null, the field selection per ticker may change overtime but I suspect it would all get routed to a same Hive datasets therefore we need to handle the evolution.

The Cassandra driver sets missing fields to null if they aren’t in the json. The question is what happens to extra fields in the inbound json which don’t exist in the table. My intended use of the Cassandra sink would be for API’s so a targeted data model. I’ll need to play with this.

PR #53, opened 2016-09-09
minor change to fix the unit test

stheppi:
@andrewstevenson : the test was failing with null pointer exception. also the create was called outside the Try so it was pointless. So now if the error happens is captured. but question is do we want to hide the exception and only log it?


Antwnis:
I'm favoring failing early on table creation issues - and logging in a banner

andrewstevenson:
@stheppi I assume this can be closed this the ReThinkSinkConverter no longer exists?

PR #65, opened 2016-10-10
Setting correct the elastic search cluster name

Antwnis:


stheppi:
I don't think this is right.
look at Sam's example for connecting:
import org.elasticsearch.common.settings.Settings
val settings = Settings.settingsBuilder().put("cluster.name", "myClusterName").build()
val client = ElasticClient.remote(settings, ElasticsearchClientUri("elasticsearch://somehost:9300"))

Antwnis:
i know thats the solution i applied

PR #66, opened 2016-10-11
tests for converting schemaless map pojo records

dollschasingmen:


dollschasingmen:
@andrewstevenson @Antwnis @stheppi

PR #68, opened 2016-10-14
supportning non avro messages

stheppi:
adding support for schemaless sink record or a  record with string schema and json payload

### DIFF
--- a/kafka-connect-influxdb/src/main/scala/com/datamountaineer/streamreactor/connect/influx/writers/InfluxBatchPointsBuilderFn.scala
+++ b/kafka-connect-influxdb/src/main/scala/com/datamountaineer/streamreactor/connect/influx/writers/InfluxBatchPointsBuilderFn.scala
@@ -19,28 +19,136 @@ package com.datamountaineer.streamreactor.connect.influx.writers
 import java.util.concurrent.TimeUnit
 
 import com.datamountaineer.streamreactor.connect.influx.config.InfluxSettings
+import com.datamountaineer.streamreactor.connect.schemas.ConverterUtil
 import io.confluent.common.config.ConfigException
-import org.apache.kafka.connect.data.Struct
+import org.apache.kafka.connect.data.{Schema, Struct}
 import org.apache.kafka.connect.sink.SinkRecord
 import org.influxdb.InfluxDB.ConsistencyLevel
 import org.influxdb.dto.{BatchPoints, Point}
 
+import scala.collection.JavaConversions._
+import scala.util.Try
 
-object InfluxBatchPointsBuilderFn {
+
+object InfluxBatchPointsBuilderFn extends ConverterUtil {
   def apply(records: Seq[SinkRecord],
             settings: InfluxSettings): BatchPoints = {
+
+    def handleSinkRecord(record: SinkRecord): Option[Point] = {
+
+      Option(record.valueSchema()) match {
+        case None =>
+          //try to take it as string
+          record.value() match {
+            case map: java.util.Map[_, _] => buildPointFromMap(record, settings)
+            case _ => sys.error("For schemaless record only String and Map types are supported")
+          }
+        case Some(schema: Schema) =>
+          schema.`type`() match {
+            case Schema.Type.STRING => buildPointFromJson(record, settings)
+            case Schema.Type.STRUCT => buildPointFromStruct(record, settings)
+            case other => sys.error(s"$other schema is not supported")
+          }
+      }
+    }
     val batchPoints = BatchPoints
       .database(settings.database)
       .tag("async", "true")
-      .retentionPolicy("default")
+      .retentionPolicy(settings.retentionPolicy)
       .consistency(ConsistencyLevel.ALL)
       .build()
 
-    records.foreach(r => buildPoint(r, settings).map(batchPoints.point))
+    records.foreach(r => handleSinkRecord(r).map(batchPoints.point))
     batchPoints
   }
 
-  private def buildPoint(record: SinkRecord, settings: InfluxSettings): Option[Point] = {
+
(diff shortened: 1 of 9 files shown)


kafka-connect-influxdb/src/main/scala/com/datamountaineer/streamreactor/connect/influx/writers/InfluxBatchPointsBuilderFn.scala
```diff
@@ -19,28 +19,136 @@ package com.datamountaineer.streamreactor.connect.influx.writers
...
             settings: InfluxSettings): BatchPoints = {
+
+    def handleSinkRecord(record: SinkRecord): Option[Point] = {
+
+      Option(record.valueSchema()) match {
+        case None =>
+          //try to take it as string
+          record.value() match {
+            case map: java.util.Map[_, _] => buildPointFromMap(record, settings)
+            case _ => sys.error("For schemaless record only String and Map types are supported")
```
Antwnis: How about Int | Long | Boolean etc ?

Antwnis: better say

`and Map[String, Any] types are supported`

kafka-connect-influxdb/src/main/scala/com/datamountaineer/streamreactor/connect/influx/writers/InfluxBatchPointsBuilderFn.scala
```diff
@@ -19,28 +19,136 @@ package com.datamountaineer.streamreactor.connect.influx.writers
...
+        case (field, value: Long) => builder.addField(field, value)
+        case (field, value: Int) => builder.addField(field, value)
+        case (field, value: BigInt) => builder.addField(field, value)
+        case (field, value: Byte) => builder.addField(field, value)
+        case (field, value: Short) => builder.addField(field, value)
+        case (field, value: Double) => builder.addField(field, value)
+        case (field, value: Float) => builder.addField(field, value)
+        case (field, value: Boolean) => builder.addField(field, value)
+        case (field, value: java.math.BigDecimal) => builder.addField(field, value)
+        case (field, value: String) => builder.addField(field, value)
```
Antwnis: shouldn't you have

```case (field, value: Map[String, Any) => ..```

To allow nested Json messages ?

File: kafka-connect-influxdb/src/test/scala/com/datamountaineer/streamreactor/connect/influx/InfluxBatchPointsBuilderFnTest.scala [hunk 3]
```diff
@@ -0,0 +1,757 @@
...
+    val map = PointMapFieldGetter.fields(point)
+    map.size shouldBe 3
+
+    map.get("_id") shouldBe "580151bca6f3a2f0577baaac"
+    map.get("this_is_renamed") shouldBe "Clements Crane"
+    map.get("email") shouldBe "clements.crane@terragen.io"
+
+  }
+
+  "throw an error of if nested json since there is no flattening of json for a sink record with string json payload" in {
```
Antwnis: I would object to that. We should provide a mechanism like

```flattenJson = '.'```

In that case we would automatically flatten the json structure using this separator

i.e. 

```  """
         | {
         |    "eyeColor": "brown",
         |    "name.first": "Christian",
         |    "name.last": "Melton"
         |    }
         | }```

File: kafka-connect-influxdb/src/test/scala/com/datamountaineer/streamreactor/connect/influx/InfluxBatchPointsBuilderFnTest.scala [hunk 4]
```diff
@@ -0,0 +1,757 @@
...
+      InfluxBatchPointsBuilderFn(Seq(record), settings)
+    }
+  }
+
+  "throw an error of if array is present in json since there is no flattening of json for a sink record with string json payload" in {
+    val jsonPayload =
+      """
+        | {
+        |    "eyeColor": "brown",
+        |     "tags": [
```
Antwnis: Here we could be flattening using the javascript style.. i.e.

```tags[0] = "ut",
     tags[1] = "dolor" ...
```

kafka-connect-influxdb/src/main/scala/com/datamountaineer/streamreactor/connect/influx/writers/InfluxBatchPointsBuilderFn.scala
stheppi: Ant, take a look at the else branch. if you don't have a schema, and you send a json and use JsonConverter you get back a java.util.Map.  the logic there is correct. If you send an int/long/Boolean what would you insert it against in influx?

kafka-connect-influxdb/src/main/scala/com/datamountaineer/streamreactor/connect/influx/writers/InfluxBatchPointsBuilderFn.scala
stheppi: What if the nested has fields already processed? basically you do an ovewrite? For now we don't support them

kafka-connect-influxdb/src/test/scala/com/datamountaineer/streamreactor/connect/influx/InfluxBatchPointsBuilderFnTest.scala
stheppi: I could do that

PR #73, opened 2016-10-19
Fix guava classloader conflict

boeboe:
Fx classloader conflict wrt guava versions 12.0.1 and 16.0.1 that conflict with a guava function used in elastic sink plugin

REF: https://github.com/datamountaineer/stream-reactor/issues/71

Within the set of connectors provided, the following guava versions are in use through transient dependencies: 12.0.1, 16.0.1, 18.0 and 19.0. The elastic sink connector depends on a guava function that was not yet available in 12.0.1 and 16.0.1... so hence the up-step of those transient dependencies.

### DIFF
--- a/kafka-connect-cassandra/build.gradle
+++ b/kafka-connect-cassandra/build.gradle
@@ -19,7 +19,11 @@ project(":kafka-connect-cassandra") {
     }
 
     dependencies {
-        compile "com.datastax.cassandra:cassandra-driver-core:$cassandraDriverVersion"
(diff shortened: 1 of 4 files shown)


andrewstevenson:
Thanks!

boeboe:
@andrewstevenson 

You need to step back the guava transient override for HBase, because it broke the functionality (hadoop-common depends on a public StopWatch constructor in Guava which has been removed (or made inaccessibly) since guava 15.0. I noticed this running the tests. For the others it should be still ok.

PR #90, opened 2016-12-12
first version

andrewstevenson:


andrewstevenson:
@stheppi We need to rebase here I think?

stheppi:
i will rebase again. need to add a few more unit tests

PR #101, opened 2016-12-29
Fix/netty native epoll on linux x86 64

andrewstevenson:


kafka-connect-kudu/src/main/scala/com/datamountaineer/streamreactor/connect/kudu/sink/KuduWriter.scala
```diff
@@ -171,19 +167,17 @@ class KuduWriter(client: KuduClient, setting: KuduSettings) extends StrictLoggin
...
       //May want to die if error policy is Throw
-      val responses = session.flush().asScala
-      if (responses != null)
-        responses
-          .filter(r => (r == null) || r.hasRowError)
-          .foreach(e => {
-            throw new Throwable(s"Failed to flush one or more changes: ${e.getRowError.toString}")
-          })
+      session.flush().asScala
+        .filter(r => Option(r).isEmpty || r.hasRowError)
```
stheppi: we removed the check for responses null. the flush can return null.

 class ConvertBatchToListOfResponsesCB implements
      Callback<Deferred<List<OperationResponse>>, ArrayList<BatchResponse>> {
    @Override
    public Deferred<List<OperationResponse>> call(ArrayList<BatchResponse> batchResponsesList)
        throws Exception {
      Deferred<List<OperationResponse>> deferred = new Deferred<>();
      if (batchResponsesList == null) {
        deferred.callback(null);
        return deferred;
      }

PR #105, opened 2016-12-30
it's non optional in the case class ..

Antwnis:
A minor fix - but did not look right


kafka-connect-blockchain/src/main/scala/com/datamountaineer/streamreactor/connect/blockchain/data/Output.scala
```diff
@@ -39,7 +39,7 @@ object Output {
     .field("addr_tag", Schema.OPTIONAL_STRING_SCHEMA)
     .field("spent", Schema.BOOLEAN_SCHEMA)
     .field("tx_index", Schema.INT64_SCHEMA)
-    .field("type", Schema.OPTIONAL_INT32_SCHEMA)
+    .field("type", Schema.INT32_SCHEMA)
```
Antwnis: line 27 is `type`: Int

PR #109, opened 2017-01-06
Adding FTP and HTTP connector

Antwnis:


Antwnis:
I'm a bit puzzled how to call this one - as this should have both FTP + HTTP capabilities
Let's keep this PR open for a while - to test and develop a bit more the connector before merging it in

PR #110, opened 2017-01-08
added support to specify the consistency level

stheppi:
removed the timer to print the stats since it wasn't thread safe (stats are still printed on 60 seconds basis)

### DIFF
--- a/kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/sink/CassandraSinkTask.scala
+++ b/kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/sink/CassandraSinkTask.scala
@@ -82,15 +82,12 @@ class CassandraSinkTask extends SinkTask with StrictLogging {
     require(writer.nonEmpty, "Writer is not set!")
     writer.foreach(w => w.write(records.toVector))
     records.foreach(r => counter.put(r.topic(), counter.getOrElse(r.topic(), 0L) + 1L))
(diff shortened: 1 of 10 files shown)


kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/sink/CassandraSinkTask.scala
```diff
@@ -87,7 +81,16 @@ class CassandraSinkTask extends SinkTask with StrictLogging {
...
     require(writer.nonEmpty, "Writer is not set!")
     writer.foreach(w => w.write(records.toVector))
-    records.foreach(r => counter.put(r.topic() , counter.getOrElse(r.topic(), 0L) + 1L))
+    records.foreach(r => counter.put(r.topic(), counter.getOrElse(r.topic(), 0L) + 1L))
+    if (timestamp > 0) {
+      val newTimestamp = System.currentTimeMillis()
+      if (scala.concurrent.duration.SECONDS.toSeconds(newTimestamp - timestamp) >= 60) {
+        logCounts()
+      }
+      timestamp = newTimestamp
```
Antwnis: Something is smelly here

It effectively does:

if ( timestamp > 0 ) timestamp = System.currentTimeMillis()

stheppi: yes. we initialize with 0 and I wanted to avoid the comparison for the first time.

PR #126, opened 2017-01-30
Add UPSERT to the ES Sink

tamarr:
Runs _update_ docAsUpsert with specified PK


kafka-connect-elastic/src/main/scala/com/datamountaineer/streamreactor/connect/elastic/ElasticJsonWriter.scala
```diff
@@ -81,7 +86,25 @@ class ElasticJsonWriter(client: ElasticClient, settings: ElasticSettings) extend
...
+                              val pkField = settings.pks(r.topic)
+                              // Extractor includes all since we already converted the records to have only needed fields
+                              val extractor = StructFieldsExtractor(includeAllFields = true, Map(pkField -> pkField))
+                              val fieldsAndValues = extractor.get(r.value.asInstanceOf[Struct]).toMap
+                              if(fieldsAndValues.contains(pkField)){
+                                val pkValue = fieldsAndValues(pkField).toString
+                                update id pkValue in i.toLowerCase / i docAsUpsert fieldsAndValues
+                              } else {
+                                logger.error(s"Empty ${pkField}")
+                                null
```
Antwnis: Should throw will a human-friendly message depending on error-policy if UPSERT is used - but no PK is defined

Antwnis: Ah sorry - it's already on the setting

andrewstevenson: Hmmm, we are applying the error policies on ES either.

tamarr: yeah, I added that because we _didn't_ want the error to explode the sink but I realise it's not your way :) I can revert that and we will add an extra check on our server side to ensure the field exist.

stheppi:
Thanks @tamarr

kafka-connect-elastic/src/main/scala/com/datamountaineer/streamreactor/connect/elastic/ElasticJsonWriter.scala
tamarr: just to clarify - the settings will throw an error if no PK is specified in the properties query, this will throw an error from ES if the PK field is missing from the record - since the id field will be missing on an update.

PR #133, opened 2017-03-02
Select column fix

CodeSmell:
Was working w/ Kafka Connect pushing data from Cassandra to a Topic
I was setting the` connect.cassandra.source.kcql` to the following `INSERT INTO idocs-topic SELECT json_doc FROM idocs_events PK idoc_event_ts` but all of the columns in the table were included in the data written to the Topic

Here is a possible fix that honors the columns in the SELECT


andrewstevenson:
@CodeSmell Thanks!

PR #139, opened 2017-03-09
Refactor to prepare for withunwrap KCQL capability

CodeSmell:
This is mostly open for comments and discussion to further explore configuration to allow more flexibility in how the `SourceRecord` is created. This is based on our conversations over email.

This PR is a possible refactor that would allow withunwrap KCQL to be plugged into the Cassandra Source. Right now the `CassandraTableReader` (line 267) has a variable `withunwrap` set to false.
The idea would be to replace this line with 

```val withunwrap = settings.withunwrap```

when the KCQL is updated and if that is the route you decide to go. 
When this is true it would create a comma delimited list out of the column values.

This allows 

`SELECT colA, ts FROM event_table IGNORE ts PK ts `

to yield a String on the Kafka Topic that contains only the value of colA

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ target/
 /kafka-connect-coap/Californium.properties
 Californium.properties
 /gradle/wrapper/gradle-wrapper.properties
+/.DS_Store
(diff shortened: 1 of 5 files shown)


andrewstevenson:
@CodeSmell Thanks!

PR #140, opened 2017-03-10
update progress counter handling

andrewstevenson:


kafka-connect-blockchain/src/main/scala/com/datamountaineer/streamreactor/connect/blockchain/source/BlockchainSourceTask.scala
```diff
@@ -79,12 +66,7 @@ class BlockchainSourceTask extends SourceTask with StrictLogging {
   override def poll(): util.List[SourceRecord] = {
     val records = blockchainManager.map(_.get()).getOrElse(new util.ArrayList[SourceRecord]())
     logger.debug(s"Returning ${records.size()} record(-s) from Blockchain source")
-    records.foreach(r => counter.put(r.topic(), counter.getOrElse(r.topic(), 0L) + 1L))
-    val newTimestamp = System.currentTimeMillis()
-    if (counter.nonEmpty && scala.concurrent.duration.SECONDS.toSeconds(newTimestamp - timestamp) >= 60) {
-      logCounts()
-    }
-    timestamp = newTimestamp
+    progressCounter.update(records.toVector)
```
stheppi: I wouldn't convert it to vector. the return is util.ArrayList and will create a new collection. the update should have a few overloaded versions maybe

kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/sink/CassandraSinkTask.scala
```diff
@@ -79,13 +74,7 @@ class CassandraSinkTask extends SinkTask with StrictLogging {
...
     require(writer.nonEmpty, "Writer is not set!")
     writer.foreach(w => w.write(records.toVector))
-    records.foreach(r => counter.put(r.topic(), counter.getOrElse(r.topic(), 0L) + 1L))
-
-    val newTimestamp = System.currentTimeMillis()
-    if (counter.nonEmpty && scala.concurrent.duration.SECONDS.toSeconds(newTimestamp - timestamp) >= 60) {
-      logCounts()
-    }
-    timestamp = newTimestamp
+    progressCounter.update(records.toVector)
```
stheppi: similar to the source. we should just pass util.collection since it has size.
I would say is better to just pass an int as count and let the caller pass the number. this way it won't matter the collection and the api stays one method

kafka-connect-blockchain/src/main/scala/com/datamountaineer/streamreactor/connect/blockchain/source/BlockchainSourceTask.scala
andrewstevenson: why won't it accept progress.update(records)?

stheppi: I it will, but because it needs to wrap the java ArrayUtils to Seq. If you do a .toVector we incur the cost of creating a Vector instance without really needing it.

andrewstevenson: intellj says no.

stheppi: you need to import scala.collection.JavaConversions._ for it to work without .toVector. but it will still wrap the underlying Java ArrayList to a Scala Seq,. It will use a JListWrapper in this case which will be cheaper than creating a  Scala vector

andrewstevenson: I have that, it doesn't like the fact it's expecting ConnectRecord but I'm giving SinkRecords.

stheppi: are the projects on same kafka lib version?

andrewstevenson: they are `0.10.1.0` and confluent `3.1.1`

PR #173, opened 2017-05-03
ElasticSearch connector with date in index and custom document type

robvadai:
The ElasticSearch connector is updated with:

- support for auto-generated time/date in index names; this is to create indexes per day instead of one ever-increasing index, as per the official guidelines on [Retiring Data](https://www.elastic.co/guide/en/elasticsearch/guide/current/retiring-data.html)
- support for custom document type for the messages to be saved; document types allow custom indexing policies

### DIFF
--- a/kafka-connect-elastic/src/main/scala/com/datamountaineer/streamreactor/connect/elastic/config/ElasticSinkConfigConstants.scala
+++ b/kafka-connect-elastic/src/main/scala/com/datamountaineer/streamreactor/connect/elastic/config/ElasticSinkConfigConstants.scala
@@ -29,4 +29,16 @@ object ElasticSinkConfigConstants {
   val URL_PREFIX_DEFAULT = "elasticsearch"
   val EXPORT_ROUTE_QUERY = "connect.elastic.sink.kcql"
   val EXPORT_ROUTE_QUERY_DOC = "KCQL expression describing field selection and routes."
+
+  val INDEX_NAME_SUFFIX = "connect.elastic.index.suffix"
+  val INDEX_NAME_SUFFIX_DOC = "Suffix to append to the index name. Supports date time notation inside curly brackets. E.g. 'abc_{YYYY-MM-dd}_def'"
+  val INDEX_NAME_SUFFIX_DEFAULT = ""
+
+  val AUTO_CREATE_INDEX = "connect.elastic.index.auto.create"
+  val AUTO_CREATE_INDEX_DOC = "The flag enables/disables auto creating the ElasticSearch index. Boolean value required. Defaults to TRUE."
+  val AUTO_CREATE_INDEX_DEFAULT = true
+
+  val DOCUMENT_TYPE = "connect.elastic.document.type"
+  val DOCUMENT_TYPE_DOC = "Sets the ElasticSearch document type. See https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-type-field.html for more info."
+  val DOCUMENT_TYPE_DEFAULT = ""
 }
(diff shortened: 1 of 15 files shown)


kafka-connect-elastic/src/main/scala/com/datamountaineer/streamreactor/connect/elastic/config/ElasticSinkConfigConstants.scala
```diff
@@ -29,4 +29,16 @@ object ElasticSinkConfigConstants {
   val URL_PREFIX_DEFAULT = "elasticsearch"
   val EXPORT_ROUTE_QUERY = "connect.elastic.sink.kcql"
   val EXPORT_ROUTE_QUERY_DOC = "KCQL expression describing field selection and routes."
+
+  val INDEX_NAME_SUFFIX = "connect.elastic.index.suffix"
```
stheppi: @robvadai 
 Shouldn't we add support for this in KCQL? The reason i am asking is for scenarios where you route messages from topic1 to index1 and topic2 to index2 and maybe you want one of the index to have suffix. Furthermore you might use different document type for messages coming from two different topics.
Thoughts?

kafka-connect-elastic/src/main/scala/com/datamountaineer/streamreactor/connect/elastic/config/ElasticSinkConfigConstants.scala
```diff
@@ -29,4 +29,16 @@ object ElasticSinkConfigConstants {
...
+  val INDEX_NAME_SUFFIX_DOC = "Suffix to append to the index name. Supports date time notation inside curly brackets. E.g. 'abc_{YYYY-MM-dd}_def'"
+  val INDEX_NAME_SUFFIX_DEFAULT = ""
+
+  val AUTO_CREATE_INDEX = "connect.elastic.index.auto.create"
+  val AUTO_CREATE_INDEX_DOC = "The flag enables/disables auto creating the ElasticSearch index. Boolean value required. Defaults to TRUE."
+  val AUTO_CREATE_INDEX_DEFAULT = true
+
+  val DOCUMENT_TYPE = "connect.elastic.document.type"
+  val DOCUMENT_TYPE_DOC = "Sets the ElasticSearch document type. See https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-type-field.html for more info."
+  val DOCUMENT_TYPE_DEFAULT = ""
```
stheppi: should probably use null and avoid toOption over an empty string

kafka-connect-elastic/src/main/scala/com/datamountaineer/streamreactor/connect/elastic/config/ElasticSinkConfigConstants.scala
stheppi: KCQL example: INSERT INTO $THE_INDEX SELECT * FROM $THE_TOPIC [WITHDOCTYPE($docType)] [WITHINDEXSUFFIX($suffix)]

robvadai: 1. Yes this could be part of KCQL, good idea, obviously updating KCQL was out of scope for me, we needed a way to support date string in ES index names
2. We had a requirement to make sure indexes are never auto created because we create indexes with custom policies and document types. The Connector was automatically creating missing indexes which is what we wanted to avoid. Hence the conditional index creation was added here: https://github.com/ConnectedHomes/stream-reactor/blob/9bc01bd61034037cacb14f5154bbfa01ae8e4b47/kafka-connect-elastic/src/main/scala/com/datamountaineer/streamreactor/connect/elastic/ElasticJsonWriter.scala#L42

stheppi:
@robvadai  thank you for the improvement

joscas:
👍

PR #176, opened 2017-05-05
Cassandra Source: use paging technique to avoid ALLOW FILTER

CodeSmell:
This is the initial outline of the approach to page through results using token of the partition key, thus enabling CQL that does not require ALLOW FILTER. 

http://docs.datastax.com/en/cql/3.1/cql/cql_using/paging_c.html

As part of this approach, moved the CQL generation to a new class

Based on the various config that affects the CQL generation, the following precedence for CQL generation modes is currently implemented:
- bulk
- INCREMENTALMODE in KCQL
- config

All existing capabilities should work even with this incomplete feature. 
There are two areas still in progress
- need to refactor `CassandraTableReader` offset so relies on `CqlGenerator` to determine type (date vs token)
- need to refactor `CassandraTableReader` to use tokenDefaultStatement vs tokenStatementWithOffset

UPDATE
- originally was using KCQL 1.0.1 but upgraded to 1.0.3

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,7 +71,7 @@ allprojects {
         json4sVersion = "3.3.0"
         kiteMiniClusterVersion = "1.1.0"
         gsonVersion = "2.6.2"
-        dataMountaineerKCQLVersion = "1.0.1"
+        dataMountaineerKCQLVersion = "1.0.3"
         akkaVersion = "2.4.10"
         connectCLIVersion = "1.0"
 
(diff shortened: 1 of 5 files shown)


CodeSmell:
Note:
One of the tests is failing at the moment. The batch size in the test of `CqlGenerator` was set to 200 but the CQL generated had the value 3000

```SELECT timestamp_field,string_field 
FROM sink_test.cassandra-table 
WHERE token(timestamp_field) > token(?) 
LIMIT 3000```

andrewstevenson:
@CodeSmell Thanks for the contribution! Can you also update the docs?

CodeSmell:
Sure, but there are couple more areas that need to be finished in the code
first.
I need to make sure the token offset is handled but right now CTReader is
assuming it is a date

PR #183, opened 2017-05-16
Fix Properties

michaelandrepearce:
Make properties create a java list of Struct using property schema.


andrewstevenson:
@michaelandrepearce Thanks!

PR #187, opened 2017-05-17
JMSSource Default Options

michaelandrepearce:
Small fix, didnt check the option on default converter essentially causing it to be mandatory when it should be optional.


michaelandrepearce:
@stheppi not sure if you seen, just a small fix to the last PR that was merged.

PR #189, opened 2017-05-24
ISSUE 179 : Support Shared Durable Subscription on JMS 2.0 Topics

michaelandrepearce:
Add configuration to provide subscription name
Bump JMS API to 2.0
Add logic to create a shared durable subscriber if destination is Topic and subscription name exists.


stheppi:
Thank you!

PR #190, opened 2017-05-24
FIX: added fetch size to Cassandra Source settings

CodeSmell:


stheppi:
@CodeSmell 
  KCQL default's 'BATCH' to 3000 (it is **Config.DEFAULT_BATCH_SIZE**). to control the fetch/batch all it needs is:
INSERT INTO kafka-topic SELECT string_field, timestamp_field FROM cassandra-table PK timestamp_field $kcqlIncrementMode **BATCH = 200**

This way you can control the batch per table rather than having a global fetch.  I am sorry for not addressing your question earlier and you had to code around it. I will have to reject the PR though.

BATCH_SIZE in the config i would say is confusing and should be removed. as records are pulled from Cassandra they are stored into a queue. This config says how many to pull at once from it. Two steps throttling.

CodeSmell:
Stefan
No problem. I think I misunderstood Andrew's email response from a few days ago. 
He mentioned tweaking `connect.cassandra.fetch.size` so I assumed that this was the config setting that should be used. 

I'll add some tests with BATCH=200 and see if that change is available in the `CqlGenerator`

Just curious what is fetch.size used for?

CodeSmell:
Based on the comment I assume the main problem is this line in `CqlGenertor`
```private val limitRowsSize = setting.fetchSize```

But would I be correct in assuming that we would still want fetch.size available in the `CassandraSetting`

PR #191, opened 2017-05-25
Add Test with KCQL Batch 

CodeSmell:
This PR is an attempt to correct #190 
It reverts the `CqlGenerator` to using the BATCH in KCQL instead of fetch.size in the config
It adds a test to show that it is working correctly w/ CQL that uses the token/paging

The `connect.cassandra.fetch.size` was added to `CassandraSettings` since it was not available to the Cassandra Source connector before. However it is currently not used in the `CqlGenerator`.

### DIFF
--- a/kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/source/CqlGenerator.scala
+++ b/kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/source/CqlGenerator.scala
@@ -23,7 +23,7 @@ class CqlGenerator(private val setting: CassandraSourceSetting) extends StrictLo
   private val keySpace = setting.keySpace
   private val selectColumns = getSelectColumns
   private val incrementMode = determineMode
-  private val limitRowsSize = config.getBatchSize
+  private val limitRowsSize = setting.fetchSize
 
   private val defaultTimestamp = "1900-01-01 00:00:00.0000000Z"
 
@@ -75,7 +75,7 @@ class CqlGenerator(private val setting: CassandraSourceSetting) extends StrictLo
     val fieldList = config.getFieldAlias.map(fa => fa.getField).toList
     // if no columns set then select all the columns in the table
     val selectColumns = if (fieldList == null || fieldList.isEmpty) "*" else fieldList.mkString(",")
-    logger.info(s"the fields to select are $selectColumns")
+    logger.debug(s"the fields to select are $selectColumns")
     selectColumns
   }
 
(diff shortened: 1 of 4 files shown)


kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/source/CqlGenerator.scala
```diff
@@ -23,7 +23,7 @@ class CqlGenerator(private val setting: CassandraSourceSetting) extends StrictLo
   private val keySpace = setting.keySpace
   private val selectColumns = getSelectColumns
   private val incrementMode = determineMode
-  private val limitRowsSize = setting.fetchSize
+  private val limitRowsSize = config.getBatchSize
```
CodeSmell: Using batch size instead of fetch size

kafka-connect-cassandra/src/test/scala/com/datamountaineer/streamreactor/connect/cassandra/source/TestCqlGenerator.scala
```diff
@@ -87,15 +87,16 @@ class TestCqlGenerator extends WordSpec with Matchers with BeforeAndAfter with M
...
       Map(
         CassandraConfigConstants.KEY_SPACE -> CASSANDRA_KEYSPACE,
         CassandraConfigConstants.SOURCE_KCQL_QUERY -> myKcql,
         CassandraConfigConstants.ASSIGNED_TABLES -> s"$TABLE3",
         CassandraConfigConstants.IMPORT_MODE -> CassandraConfigConstants.INCREMENTAL,
         CassandraConfigConstants.POLL_INTERVAL -> "1000",
-        CassandraConfigConstants.FETCH_SIZE -> "200",
+        CassandraConfigConstants.FETCH_SIZE -> "500",
+        CassandraConfigConstants.BATCH_SIZE -> "800",
         CassandraConfigConstants.TIMESTAMP_TYPE -> configIncrementMode).asJava
```
CodeSmell: set the values for FETCH & BATCH size in the test here w/ numbers different than KCQL to test/confirm that the KCQL value is being used

stheppi:
I will do some cleanup . Fetch.size should be removed and batch size is confusing

CodeSmell:
Thanks. I have a better understanding now. 
But having batch in KCQL, fetch.size config, and batch.size config was confusing. :)

andrewstevenson:
We are cleaning up all the configs. Spotted a few more today. Are you deploying into a containers engines at all?

CodeSmell:
We are deploying into a private OneOps/OpenStax cloud.
KC is dropped on own nodes with Stream Reactor jar dropped into the libs.

PR #211, opened 2017-06-02
Cleaned up token test and added a warning

CodeSmell:
The TOKEN increment mode only works as expected when the ByteOrderedPartitioner is being used. 
This is generally not recommended due to hotspots but in certain cases might be helpful to avoid ALLOW FILTERING and the lack of a LIMIT on time based polling of Cassandra table. 

The available partitioners:
http://docs.datastax.com/en/cassandra/3.0/cassandra/architecture/archPartitionerAbout.html
https://wiki.apache.org/cassandra/Partitioners

It can be used with other partitioners but inserts can be missed that occur after a query is run because other partitioners have no order in the way tokens are ordered.


CodeSmell:
Note: after doing a pull from upstream had trouble w/ some of the temp classes. 
So added this in the older branch.

CodeSmell:
Note: after doing a pull from upstream into fork got some errors with missing temp classes.
So added these changes in the older branch

stheppi:
@CodeSmell 
 can you please resolve the merge conflicts so i can merge?

stheppi:
Thanks!

PR #227, opened 2017-07-12
Feature/kcql 2.0

stheppi:
adding support for nested fields in cassandra sink
added support for inserting to multiple column families from the same kafka topic

### DIFF
--- a/kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/sink/Transform.scala
+++ b/kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/sink/Transform.scala
@@ -1,0 +1,107 @@
+package com.datamountaineer.streamreactor.connect.cassandra.sink
+
+import java.nio.ByteBuffer
+
+import com.datamountaineer.streamreactor.connect.json.SimpleJsonConverter
+import com.fasterxml.jackson.databind.JsonNode
+import com.landoop.connect.sql.StructSql._
+import com.landoop.json.sql.JsonSql._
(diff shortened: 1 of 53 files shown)


kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/sink/Transform.scala
```diff
@@ -0,0 +1,107 @@
...
+                }
+            }
+          //we take it as String
+          case other => raiseException(s"Value:$other is not handled!", null)
+        }
+      }
+    }
+  }
+}
+
```
andrewstevenson: Will this not be common across many connectors?

kafka-connect-cassandra/src/test/resources/output1.json
```diff
@@ -0,0 +1,11 @@
+{
+  "addr_tag_link": null,
+  "addr_tag": null,
```
andrewstevenson: Are these meant to be checked in?

stheppi: Yes. I think it should

kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/sink/Transform.scala
stheppi: It will be shared for json transformations

andrewstevenson: So shouldn't we move it to Common?

PR #248, opened 2017-08-09
fixing the namespaces

stheppi:


stheppi:
I am not sure about consolidating into one. We use different elastic search dependencies. Jvm jar nightmare better be avoided

stheppi:
Users should get the version they use 2 or 5. Why would they get and try both?

Antwnis:
wget stream-reactor.gz 

^ they get all connectors and just unzip them 

Similarly fast-data-dev docker contains all connectors on the classpath

stheppi:
I disagree with you to handle ES5 and ES2 into the same package. TCP connection libs comes with different versions; putting all jars in the same path for two different versions is a recipe for problems.

Issue #267, opened 2017-09-06
High CPU on MQTT source


masahirom:
Enabling mqtt-source-connector causes high CPU usage even when there is no message published. 
One active connector spends 100% CPU, and two active connectors spend 200% CPU.

Version: stream-reactor-0.2.5-3.2.2

Here is my configuration.

```
name=mqtt-source-json
tasks.max=1
connect.mqtt.connection.clean=true
connect.mqtt.connection.timeout=3000
connect.mqtt.source.kcql=INSERT INTO test SELECT * FROM mqtt_test
connect.mqtt.connection.keep.alive=5000
connect.mqtt.source.converters=mqtt_test=com.datamountaier.streamreactor.connect.converters.source.JsonSimpleConverter
connect.mqtt.client.id=mqtt-source-json-test
connect.mqtt.converter.throw.on.error=true
connect.mqtt.hosts=tcp://127.0.0.1:1883
connect.mqtt.service.quality=1
connector.class=com.datamountaineer.streamreactor.connect.mqtt.source.MqttSourceConnector
```

The method getRecords() in MqttManager.scala seems to be called by poll() in MqttSourceTask.scala with no wait.
I modified the getRecords() to poll the internal queue with timeout, it looks fine.

Original : com/datamountaineer/streamreactor/connect/mqtt/source/MqttSourceTask.scala
```
def getRecords(target: util.Collection[SourceRecord]): Int = queue.drainTo(target)
```

Modified
```
def getRecords(target: util.Collection[SourceRecord]): Int = {
    target.add(queue.take())
    queue.drainTo(target) + 1
}
```
Side effect: Connector's taskState remains RUNNING when we pause connector with no incomming message.

So, below may be better.
```
def getRecords(target: util.Collection[SourceRecord]): Int = {
    Option(queue.poll(1000, TimeUnit.MILLISECONDS)) match {
        case Some(x) =>
            target.add(x)
            queue.drainTo(target) + 1
        case None =>
            0
    }
}
```

andrewstevenson:
closed by pr.

PR #268, opened 2017-09-06
fixed: high cpu on mqtt-source

masahirom:
High CPU on MQTT source #267 

added configuration option: 'polling.timeout'


stheppi:
Awesome. Thank you.

andrewstevenson:
@masahirom Do you mind telling us about your use case?

masahirom:
Thank you for merging.
That is not for a specific use case. We are now doing performance evaluation for research.

PR #280, opened 2017-09-20
Fix for retrieving message data

pliew:
Fix for conversion of Bloomberg Datetime to OffsetTime (Bloomberg datetime objects represent values from 1-24 as opposed to java 0-23)
Fixed element traversal

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/280)

### DIFF
--- a/kafka-connect-bloomberg/src/main/scala/com/datamountaineer/streamreactor/connect/bloomberg/BloombergFieldValueFn.scala
+++ b/kafka-connect-bloomberg/src/main/scala/com/datamountaineer/streamreactor/connect/bloomberg/BloombergFieldValueFn.scala
@@ -84,7 +84,7 @@ private[bloomberg] object BloombergFieldValueFn {
       0
     }
     val offset = ZoneOffset.ofTotalSeconds(offsetSeconds)
-    OffsetTime.of(dt.hour(), dt.minute(), dt.second(), dt.nanosecond(), offset).toString
+    OffsetTime.of(dt.hour()-1, dt.minute(), dt.second(), dt.nanosecond(), offset).toString
   }
 
   def localDate(dt: Datetime): String = {
--- a/kafka-connect-bloomberg/src/main/scala/com/datamountaineer/streamreactor/connect/bloomberg/BloombergData.scala
+++ b/kafka-connect-bloomberg/src/main/scala/com/datamountaineer/streamreactor/connect/bloomberg/BloombergData.scala
@@ -47,7 +47,7 @@ private[bloomberg] object BloombergData {
     * @return
     */
   def apply(ticker: String, element: Element): BloombergData = {
-    val fields = (0 until element.numValues())
+    val fields = (0 until element.numElements())
       .map(element.getElement)
       .filter(f => !f.isNull)
       .foldLeft {


kafka-connect-bloomberg/src/main/scala/com/datamountaineer/streamreactor/connect/bloomberg/BloombergFieldValueFn.scala
```diff
@@ -84,7 +84,7 @@ private[bloomberg] object BloombergFieldValueFn {
       0
     }
     val offset = ZoneOffset.ofTotalSeconds(offsetSeconds)
-    OffsetTime.of(dt.hour(), dt.minute(), dt.second(), dt.nanosecond(), offset).toString
+    OffsetTime.of(dt.hour()-1, dt.minute(), dt.second(), dt.nanosecond(), offset).toString
```
stheppi: i don't think is right. are you adjusting because of bloomberg works on GMT?  dt should be in GMT, is it not?

pliew: I agree.  This one didn't seem right with me either, but when running it I noticed that dt.hour() would return 24.  This shouldn't be possible regardless of GMT or not.  There documentation makes no mention of whether it is 0 based or 1 based.
The scala code seems to mirror the jbloomberg utils and you would think they would have seen this issue as well.

Antwnis: https://rdrr.io/cran/Rblpapi/man/getTicks.html
`startTime - A Datetime object with the start time, defaults to one hour before current time`

What i'd suggest is to make this -1 as a configurable thing and have it default to -1
There is always the chance that your are getting a conversion between UTC and BST (British Summer Time) - and without me knowing more, maybe on some other regions this scew is different ..?

andrewstevenson:
@pliew I can add this in if you prefer?

pliew:
Hey Andrew,
I can add it in.

Antwnis:
There is a tiny bit of work to be done to merge this PR. Let's do it 👍 and bring it in plz

kafka-connect-bloomberg/src/main/scala/com/datamountaineer/streamreactor/connect/bloomberg/BloombergFieldValueFn.scala
assylias: `dt.hour() > 23` can only happen when a field has `dt.hasParts(DateTime.TIME)` set to false - so you shouldn't be using the hour field in that case. Your issue is due to Bloomberg sometimes returning (wrongly IMO) a `Dataype.TIME` datatype although the field has `dt.hasParts(DateTime.TIME)` set to false - in other words the field is flagged as a `TIME` but it's a `DATE`. One way to fix it is to change the logic in `BloombergFieldValueFn` where you would check the presence of the date and time parts regardless of the type (DATE, TIME or DATETIME) and call the appropriate method. BTW a time only field is possible.  
FYI Bloomberg say that it's by design so they won't fix the behaviour.

andmarios:
testit

PR #285, opened 2017-09-28
JMS source session not transacted so that client acking works.

### LINKED ISSUE #226: Kafka Connect JMS Source-messages not dequeued
Hi I followed the steps described here http://docs.datamountaineer.com/en/latest/jms-source.html
I could see and receive the messages with kafka-avro-console-consumer, however when I checked the activemq admin web console the messages are not being dequeued. It looks like for some reason the messages are not being acknowledged.

matthedude:
Previously the JMS Source was set to use a transacted session which caused the acknowledge mode to be ignored.

Fixes #226


stheppi:
thank you!

PR #288, opened 2017-10-02
JMS Acknowledge after messages have been processed by Kafka Connect

matthedude:
This PR moves the acking of messages to the `commit` interface of a Kafka Connect `SourceTask`. From [here](https://docs.confluent.io/current/connect/devguide.html#task-example-source-task) it appears that acking in the `commit` will give us at-least-once semantics.


matthedude:
NB Only 1 message is acked in `commit` as with Client Acknowledge all the in flight messages are acked by making the call on a single message.

PR #292, opened 2017-10-05
Fix: Improve JMS acking test

matthedude:
Previously we were not checking for prematurely acked messages (meaning someone could change to Auto-Ack and the tests would pass).


andrewstevenson:
Thanks! @matthedude

PR #296, opened 2017-10-15
Use a map to represent the properties field in the jms avro schema

lacarvalho91:
An optional array where each element has an optional key and an optional value seems to make more sense as an optional Map. 

Could even make the map empty if there are no properties instead of making the Map optional - not entirely sure of the consequences of that though for users of the schema. 

What do you guys think?


andrewstevenson:
@lacarvalho91 thanks!

PR #313, opened 2017-11-01
1.Add authentication 2.Use general subscription string 3.Include serv…

qianbo54321:
…ice topic in subscription string 4.Fix bug in BloombergData and BloombergFieldValueFn apply functions.


stheppi:
awesome, thanks!

PR #327, opened 2017-11-22
fix(kafka-connect-jms): ActiveMQ libs are only required at test-time

### LINKED ISSUE #326: activemq library used for testing present in kafka-connect-jms fat jar
The ActiveMQ library is only required at test time. Leaving it in compile-time scope puts it into the Shadow/Fat Jar which isn't necessary. Users of the fat jar can use any implementation of a JMS broker (e.g. WebSphereMQ, RabbitMQ etc), so bundling the ActiveMQ jar is a bit misleading. Also, the ActiveMQ jar that does get bundled is over 5 years old (v5.7) so likely contains numerous bugs now (the latest version is v5.15 - http://activemq.apache.org/download.html).

richardlundsky:
Change scope of dependency from 'compile' to 'testCompile'

Fixes #326

### DIFF
--- a/kafka-connect-jms/build.gradle
+++ b/kafka-connect-jms/build.gradle
@@ -23,8 +23,8 @@ project(":kafka-connect-jms") {
     }
 
     dependencies {
-        compile("org.apache.activemq:activemq-core:5.7.0")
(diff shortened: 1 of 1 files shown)


stheppi:
@richardlundsky thank you for contributing.

PR #331, opened 2017-11-24
Added support for filtering files via RegEx

Tribwerk:
Most of the times you don't want all files in a certain directory to be processed. Added support for filtering files via RegEx pattern to the FTP Connector. Makes it possible to tell the connector to select files that only match a certain pattern set in the config. Default behaviour is matching all.


Tribwerk:
FTP Connector

PR #343, opened 2017-12-14
bugs: Fix the List<UDT> serialization in the cassandra connector.

AERC18:


stheppi:
@AERC18 thanks for making the connector better.

PR #353, opened 2018-01-09
Fix JMS connector message acking.

matthedude:
There is a problem using the Kafka Connect `commit()` method detailed [here](https://issues.apache.org/jira/browse/KAFKA-5716). We observed messages being acked when there was a task failure (eg. schema registry being unavailable) even though they had not been processed.

We have followed a suggested fix which is to override the more granular `commitRecord(SourceRecord record)` method and keep track of records that need to be committed.


andrewstevenson:
@matthedude Awesome. Thanks!

PR #356, opened 2018-01-17
add pulsar-checksum-1.20.0-incubating jar to dependencies

andmarios:
Signed-off-by: Marios Andreopoulos <opensource@andmarios.com>


andmarios:
lanbot testit

PR #363, opened 2018-01-18
Close connection to broker when shutting down a task.

matthedude:
Previously the JMS connector was not closing the connection to the broker when a source task was shut down which could cause issues when reconnecting.

### DIFF
--- a/kafka-connect-jms/src/main/scala/com/datamountaineer/streamreactor/connect/jms/JMSSessionProvider.scala
+++ b/kafka-connect-jms/src/main/scala/com/datamountaineer/streamreactor/connect/jms/JMSSessionProvider.scala
@@ -45,29 +45,11 @@ case class JMSSessionProvider(queueConsumers: Map[String, MessageConsumer],
     connection.start()
   }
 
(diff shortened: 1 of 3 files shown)


kafka-connect-jms/src/main/scala/com/datamountaineer/streamreactor/connect/jms/JMSSessionProvider.scala
```diff
@@ -45,29 +45,11 @@ case class JMSSessionProvider(queueConsumers: Map[String, MessageConsumer],
     connection.start()
   }
 
-  def close(): Try[Unit] = {
```
stheppi: does it release the consumers/producers automatically when connection/context is closed?

matthedude: I tried closing the comsumers/producers after closing the connection and it said they were already closed 👍

stheppi:
@matthedude  thank you!

PR #391, opened 2018-02-06
Connector DELETE in waiting interval step

caiooliveiraeti:
When connector is in waiting interval step and you send "DELETE /connectors". Kafka connect worker has problem for stop thread.

And when connector is paused state and you send "DELETE /connectors". Kafka connect worker has problem for stop thread.

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/391)

### DIFF
--- a/kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/source/CassandraTableReader.scala
+++ b/kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/source/CassandraTableReader.scala
@@ -329,7 +329,12 @@ class CassandraTableReader(private val name: String,
     }
 
     // add source record to queue
-    while (!queue.offer(record, 1, TimeUnit.SECONDS)) {
+    var shouldStop = false
+    while (!queue.offer(record, 1, TimeUnit.SECONDS) & !shouldStop) {
+      shouldStop = stop.get()
+    }
+    if(shouldStop){
+      throw new ConnectException(s"Connector $name stopped for table $keySpace.$table.")
     }
   }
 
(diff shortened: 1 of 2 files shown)


kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/source/CassandraTableReader.scala
```diff
@@ -329,7 +329,12 @@ class CassandraTableReader(private val name: String,
     }
 
     // add source record to queue
-    while (!queue.offer(record, 1, TimeUnit.SECONDS)) {
+    var shouldStop = false
```
CodeSmell: this might be cleaner as
```
while (!queue.offer(record, 1, TimeUnit.SECONDS) & !stop.get()) {
}
```
is there any reason to throw the exception?

caiooliveiraeti: I check if is stopped only when have timeout for putting in queue. So I dont need get volatile value to queue all rows.

When task is stopped I need stop a process. For stop a reader

CodeSmell: this method only processes 1 row, so (unless missing something) 
if the `offer` is successful it would stop looping and exit the method
if the `offer` is not successful after 1 second (ie queue is full) it will check the stop
if the stop is false it will process loop again 
if the stop is true it will stop looping and exit the method

if we don't throw the exception here `processRow` exits and then the method that called it detects the stop and handles it

caiooliveiraeti: If we don't throw the exception the readers can wait 100 seconds for stop. 

This because a ResultSet loop will check if is stopped only each 100 rows.  Because this I added a variable for control. 

This variable start false and check if stopped only if queue is full. I tried using like your suggestion and needed 100 seconds for my connector worker works again.

With my code needed only 1 seconds for stop when is paused.

Antwnis: In the light of this discussion and of the blog article at https://theagilejedi.wordpress.com/2018/02/12/tuning-the-kafka-connect-cassandra-source-part-2/ how shall we progress with this PR @CodeSmell how does it look ?

PR #394, opened 2018-02-07
Fix bigdecimal with scale greater than 18

caiooliveiraeti:
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/394)


andmarios:
testit

PR #395, opened 2018-02-08
Closing JMS connection on failing to instantiate JMS session provider

mludwinowiczsky:
The JMS connection was never closed when an exception was thrown when instantiating JMS connection provider (e.g. configuring producers or consumers). It was just hanging forever, not allowing to create a new connection with the same id and unnecessarily using ActiveMQ resources.
This PR fixes the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/395)


stheppi:
@mludwinowiczsky 
Nice fix!
Thank you.

PR #398, opened 2018-02-14
for connect framework to correctly resolve if it is sink or source

Antwnis:
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/398)


stheppi:
Nope. Thr code does a check on isAssignable from the interfaces SourceConnector SinkConnector

andrewstevenson:
@Antwnis so renaming the class isn't the cause?

PR #412, opened 2018-03-19
Fix/cassandra source state corrupted

stheppi:
THere are still room for improvements there.

Here are areas where this should be improved,For example:
In the Task  poll method it iterates through all the tables and for each one it checks to see if it should query and if not it waits for the interval thus it delays (potentially) the query for the second table in the KCQL list

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/412)


stheppi:
@CodeSmell please review this fix.

kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/source/CassandraTableReader.scala
```diff
@@ -340,9 +344,6 @@ class CassandraTableReader(private val name: String,
     while (!queue.offer(record, 1, TimeUnit.SECONDS) & !shouldStop) {
       shouldStop = stop.get()
     }
-    if(shouldStop){
-      throw new ConnectException(s"Connector $name stopped for table $keySpace.$table.")
```
caiooliveiraeti: @stheppi,if remove this, a connector can wait 100 seconds for stop. It is for force stop. This exception is catched in [this line](https://github.com/Landoop/stream-reactor/blob/92987296ae78774befa6bc56e6c12b71de6b2a3a/kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/source/CassandraTableReader.scala#L249)

stheppi: i pushed a change.
the offer was called first before the check on the stop

now that is avoid. previousely the loop was stop by an exception. not an ideal way breaking the loop

kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/source/CassandraTableReader.scala
```diff
@@ -336,13 +331,11 @@ class CassandraTableReader(private val name: String,
     }
 
     // add source record to queue
-    var shouldStop = false
-    while (!queue.offer(record, 1, TimeUnit.SECONDS) & !shouldStop) {
-      shouldStop = stop.get()
-    }
-    if(shouldStop){
-      throw new ConnectException(s"Connector $name stopped for table $keySpace.$table.")
+    var added = false
```
CodeSmell: nice. much cleaner

kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/source/CassandraTableReader.scala
caiooliveiraeti: About valatile cache. Don't have problem remove?

kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/source/CassandraTableReader.scala
```diff
@@ -213,65 +213,60 @@ class CassandraTableReader(private val name: String,
...
+                logger.debug(s"Connector $name max Offset is currently: ${maxOffset.get}")
+              }
+              if(processRow(row)) {
+                counter += 1
               }
-              logger.debug(s"Connector $name max Offset is currently: ${maxOffset.get}")
+            } match {
+              case Failure(e) =>
+                logger.error(s"Connector $name error processing row ${row.toString} for table $keySpace.$table.", e)
+                throw new ConnectException(s"Connector $name error processing row ${row.toString} for table $keySpace.$table.", e)
```
CodeSmell: trying to follow code flow
is the `case Failure(e)` catching the exception from `processRow` but then throwing a `ConnectException`  that we recatch below?

stheppi: nice catch. 
i just cleaned that code. The Try block was not adding any value.

what we don't do is propagate the exception to connect to retry. after this PR maybe we address that.

kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/source/CassandraTableReader.scala
stheppi: it will affect the code a bit in terms of performance. but we check the variable in the processRow anyway. so not much reason to cache it then

CodeSmell:
LGTM

PR #425, opened 2018-04-18
Implemented resubscribe on reconnect logic

### LINKED ISSUE #419: MQTT Source connector does not re-subscribe on reconnect
The MQTT Source connector uses the paho client's built-in mechanism for reconnecting when the connection to the broker is lost.

> In the event that the connection is lost, the paho client will attempt to reconnect to the server. It will initially wait 1 second before it attempts to reconnect, for every failed reconnect attempt, the delay will double until it is at 2 minutes at which point the delay will stay at 2 minutes.
> Source: https://github.com/eclipse/paho.mqtt.java/blob/v1.2.0/org.eclipse.paho.client.mqttv3/src/main/java/org/eclipse/paho/client/mqttv3/MqttConnectOptions.java#L593

When the broker restarts and sessions are not persisted, subscriptions of the connector are lost. If sessions are persisted, the subscription **might** still be present; if the broker loses the entire state during restart then all subscriptions are lost regardless of session persistence.

NiekPeeters:
Fixes #419

- Refactored MqttManager: client is now created by MqttManager instead of MqttClientConnectionFn (necessary because callback is invoked before MqttClientConnectionFn returned the MqttClient, and subscribing moved to the onConnect callback.
- Changes to MqttSourceTask: no more implicit MqttSourceSettings: now explicitly passed to MqttManager instead of MqttClientConnectionFn
- Changes to MqttManagerTest: Added try-finally block so MqttManager is closed even when tests fail. Added test that verifies that manager resubscribes after reconnect.

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/425)

### DIFF
--- a/kafka-connect-mqtt/src/main/scala/com/datamountaineer/streamreactor/connect/mqtt/connection/MqttClientConnectionFn.scala
+++ b/kafka-connect-mqtt/src/main/scala/com/datamountaineer/streamreactor/connect/mqtt/connection/MqttClientConnectionFn.scala
@@ -23,9 +23,9 @@ import org.eclipse.paho.client.mqttv3.persist.MemoryPersistence
 import org.eclipse.paho.client.mqttv3.{MqttCallback, MqttClient, MqttConnectOptions}
 
 object MqttClientConnectionFn extends StrictLogging {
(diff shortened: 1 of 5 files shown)


andrewstevenson:
@NiekPeeters thanks!

PR #454, opened 2018-06-04
Bugfix for detecting Sorted Set mode

Rdebokx:
**Problem description:**
When using a KCQL statement that uses the Sorted Set mode, this is not being picked up by the RedisSinkTask, more specifically, not by this check: https://github.com/Landoop/stream-reactor/blob/master/kafka-connect-redis/src/main/scala/com/datamountaineer/streamreactor/connect/redis/sink/RedisSinkTask.scala#L75

This seems to be due to the fact that the primary keys array is an empty array, rather than it being null when no primary keys are provided in the KCQL statement. This results in the application throwing an exception stating that no writers were set per https://github.com/Landoop/stream-reactor/blob/master/kafka-connect-redis/src/main/scala/com/datamountaineer/streamreactor/connect/redis/sink/RedisSinkTask.scala#L100

This PR includes a patch for this issue, where the primary key array will be checked for its length rather than it being null, as well as Scala Tests to verify that this patch does indeed resolve aforementioned problem.

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/454)


andrewstevenson:
@Rdebokx thanks!

Issue #470, opened 2018-06-23
Cassandra Sink handles incorrectly struct key types while deleting row


piotrekfus91:
Cassandra is a database which is very strict in checking types. If we give `int` while the column is defined as `bigint` it fails with `CodecNotFoundException`.

If key is of `struct`, key fields are not extracted directly, but using JSON conversion (https://github.com/Landoop/stream-reactor/blob/master/kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/utils/KeyUtils.scala#L31). The problem is that JSON library deserializes numeric type to a type which is chosen based on the value. It completely ignores schema types (short, int, etc). If field is defined in schema as `INT64` and we have value which is greater than `Integer.MAX_INT` - it works ok, because JSON deserializes `Long`. But if using same schema definition value is smaller than `Integer.MAX_INT`, the value is cast to `Integer`, which breaks deletion using exception above.

Reproduction:
1. Create table with bigint as key (say `key1`).
2. Randomize `key1` as long greater than `Integer.MAX_INT`.
3. Insert record to Cassandra.
4. Send delete message, using `key1` and `STRUCT` schema.
5. Check if row is absent in Cassandra - should work.
6. Randomize `key` as Long, but of value smaller than `Integer.MAX_INT`.
7. Insert record to Cassandra.
8. Send delete message, using new `key1` and `STRUCT` schema.
9. Exception is thrown.

davidsloan:
Seeing as your PR has gone in I hope you don't mind if I close this, as there doesn't seem to be anything outstanding here.  Let me know if I am mistaken.

PR #471, opened 2018-06-23
Use correct field types while extracting from struct key (#470)

piotrekfus91:
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/471)


andrewstevenson:
@piotrekfus91 Thanks!

kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/utils/KeyUtils.scala
```diff
@@ -28,6 +27,14 @@ object KeyUtils {
     * @return
     */
   def keysFromStruct(struct: Struct, schema: Schema, fieldNames: Seq[String]): Seq[Object] =
-    keysFromJson(new SimpleJsonConverter().fromConnectData(schema, struct).toString, fieldNames)
+    fieldNames.map(getKeyFromStruct(struct, _))
```
stheppi: we have 
`com.datamountaineer.streamreactor.connect.converters.source.KeyExtractor::extract(Struct, Vector[String])`

if you split by `.` you have the vector[string] and the struct is there already.

PR #487, opened 2018-07-31
Support primitive types in `connect.cassandra.delete.struct_flds`

v-gerasimov:
At this moment there is no way to use primitive types in `connect.cassandra.delete.struct_flds` if schema type is `Schema.Type.STRING`. Also we cannot use `connect.cassandra.delete.struct_flds` without definition if `connect.cassandra.delete.statement` was defined because we got `InvalidQueryException: Invalid unset value for column` at runtime. I suggest to use default value of `connect.cassandra.delete.struct_flds` for primitive types. On the other hand if we define `connect.cassandra.delete.struct_flds` it represents string key as JSON.

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/487)


stheppi:
Thanks for contributing!

PR #510, opened 2018-11-14
fix bug: creating kudu table multiple times

libenchao:
in `com.datamountaineer.streamreactor.connect.kudu.sink.DbHandler`:createTables, there is no need for the outer loop on `subjects`. And the loop on `subjects` will lead to creating same table multiple times, which will be failed for the second time.

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/510)


andrewstevenson:
@libenchao Thanks!

PR #515, opened 2018-12-13
Support for Cassandra password file

alfredo-gimenez:
For security/infrastructure reasons we've opted for using password files rather than providing passwords as strings. This implementation is fairly straightforward, though it does prefer using a password file if both are provided.

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/515)

### DIFF
--- a/kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/CassandraConnection.scala
+++ b/kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/CassandraConnection.scala
@@ -24,6 +24,8 @@ import com.datastax.driver.core.{Cluster, JdkSSLOptions, QueryOptions, Session}
 import com.typesafe.scalalogging.slf4j.StrictLogging
 import org.apache.kafka.common.config.AbstractConfig
 
+import scala.io.Source
+
 /**
   * Set up a Casssandra connection
   **/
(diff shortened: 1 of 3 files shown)


andrewstevenson:
@alfredo-gimenez Does Kafka Connect external secrets not solve your issue? https://cwiki.apache.org/confluence/display/KAFKA/KIP-297%3A+Externalizing+Secrets+for+Connect+Configurations

alfredo-gimenez:
I didn't know about this, but it's exactly what we need, thanks for the pointer! 

If I'm understanding this correctly, we can do (in Kafka 2.0.0+) something like:

```
config.providers=file
config.providers.file.class=org.apache.kafka.connect.configs.FileConfigProvider
config.providers.file.param.path=/my/secret/password/file.txt

connect.cassandra.password=${file:cassandra_password}
```

with `/my/secret/password/file.txt`:

```
cassandra_password=hunter2
```

And as long as the source/sink uses the new `Map<String, String> configs();` in Source/SinkTaskContext, the indirection will be resolved at runtime (which is excellent). 

However, maybe I'm missing something but it doesn't look like the current version actually uses the new `configs()` function. Here: https://github.com/Landoop/stream-reactor/blob/master/kafka-connect-cassandra/src/main/scala/com/datamountaineer/streamreactor/connect/cassandra/sink/CassandraSinkTask.scala#L51 should be patched like:

```
- val taskConfig = Try(new CassandraConfigSink(props))
+ val taskConfig = Try(new CassandraConfigSink(context.configs()))
```
correct?

andrewstevenson:
@alfredo-gimenez I'll take a look.

andrewstevenson:
@alfredo-gimenez #526 should cover this. I'm closing this, reopen if it doesn't work for you

PR #523, opened 2019-01-04
fix records out of order when insert

sdlcwangsong:
When I insert into data to kudu, I got some out of order records in kudu, and I thought this is a bug

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/523)


andrewstevenson:
@sdlcwangsong Thanks!

PR #529, opened 2019-01-17
fix(kafka-connect-cassandra): Set support on Cassandra Source Connector

sotirissl:
- Fixed Set support on Cassandra Source Connector
- One test case added

Fixes #492

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/529)

PR #530, opened 2019-01-17
Feature/517

sotirissl:
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/530)


Antwnis:
Did we close both #529 and #530 @sotirissl ?

PR #541, opened 2019-02-04
Adds ttl support to redis

masihyeganeh:
It is crucial in my use case for Redis keys to have TTL, so I add it.
I'm not sure if it is the right way or not. but please merge it if it is.

Thanks

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/541)


Kanris826:
It would be good to see how to use this feature in README.md

PR #557, opened 2019-03-13
Set http basic auth user/password to empty strings

Thor77:
to prevent NullPointerException when using HTTP without setting
authentication credentials.

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/557)


andrewstevenson:
@Thor77 Thanks!

PR #565, opened 2019-04-09
fix: adds compatibility with messages of type com.tibco.tibjms.Tibjms…

hectorgarzon:
…ObjectMessage

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/565)


andrewstevenson:
@hectorgarzon I'm going to merge this into master. We generally don't merge into older releases and only fix forward.

PR #580, opened 2019-05-06
Featrue/fix cannot add columns

sdlcwangsong:
Hi, I found a bug about adding columns:
    If i add some columns in my avro schema, and doesn't send any record by this new schema, then I restart kafka-connect-kudu task.
    After the task is restarted, I send some record by this new schema, however, the new columns are not added to the kudu table. 
   As we know, the condition of evolution is to judge  whether the schema version is changed or not，and when we restart this task，we will get the same schema version always.I change the condition of evolution to  judge whether the current record schema columns' size is larger than kudu table columns' size or not.

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/580)

### DIFF
--- a/kafka-connect-kudu/src/main/scala/com/datamountaineer/streamreactor/connect/kudu/sink/KuduWriter.scala
+++ b/kafka-connect-kudu/src/main/scala/com/datamountaineer/streamreactor/connect/kudu/sink/KuduWriter.scala
@@ -187,16 +187,21 @@ class KuduWriter(client: KuduClient, setting: KuduSettings) extends StrictLoggin
       val schema = record.valueSchema()
       val version = schema.version()
       val table = setting.topicTables(topic)
-      val cachedSchema = schemaCache.getOrElse(topic, SchemaMap(version, schema))
 
+      var currentColumnsSize = record.valueSchema().fields().size()
+      var oladKuduTableSchema = client.openTable(table).getSchema
+      var oldColumnsSize = oladKuduTableSchema.getColumns.size()
       //allow evolution
-      val evolving = cachedSchema.version < version
+      val evolving = oldColumnsSize < currentColumnsSize
 
+      var kcql = setting.kcql.filter(r => r.getTarget.trim == table)
+      var currentKuduTableSchema = convertToKuduSchema(record, kcql.head)
       //if table is allowed to evolve all the table
       if (evolving) {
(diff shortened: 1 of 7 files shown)


kafka-connect-kudu/src/main/scala/com/datamountaineer/streamreactor/connect/kudu/sink/KuduWriter.scala
```diff
@@ -187,16 +187,23 @@ class KuduWriter(client: KuduClient, setting: KuduSettings) extends StrictLoggin
       val schema = record.valueSchema()
       val version = schema.version()
       val table = setting.topicTables(topic)
-      val cachedSchema = schemaCache.getOrElse(topic, SchemaMap(version, schema))
+
+      var currentKcql = setting.kcql.filter(r => r.getTarget.trim == table)
```
andrewstevenson: @sdlcwangsong Why are these variables `var` not `val`?

kafka-connect-kudu/src/main/scala/com/datamountaineer/streamreactor/connect/kudu/sink/KuduWriter.scala
```diff
@@ -187,16 +187,23 @@ class KuduWriter(client: KuduClient, setting: KuduSettings) extends StrictLoggin
...
+      var currentKuduTableSchema = convertToKuduSchema(record, currentKcql.head)
+      var currentColumnsSize = record.valueSchema().fields().size()
+
+      var oladKuduTableSchema = kuduTablesCache(topic).getSchema
```
andrewstevenson: looks like a type in the name

sdlcwangsong: sorry, my fault

kafka-connect-kudu/src/main/scala/com/datamountaineer/streamreactor/connect/kudu/sink/KuduWriter.scala
sdlcwangsong: ok, I will modify var  to val

sdlcwangsong:
I will add as soon as possible write unit test

sdlcwangsong:
hi, @andrewstevenson,
 I modify unit test, and fix word typo.

PR #593, opened 2019-07-05
Added support for Redis GEOADD command.

hsomyk:
Implemented new RedisGeoAdd writer to add support for Redis GEOADD command.
Added unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/593)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -117,6 +117,7 @@ allprojects {
         maven {
             url 'http://dl.bintray.com/andsel/maven/'
         }
+        maven { url "https://dl.bintray.com/spinnaker/spinnaker/" }
     }
 
     jar {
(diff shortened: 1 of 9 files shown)


andrewstevenson:
@hsomyk The build fails? Are you adding the correct jedis version? https://circleci.com/gh/Landoop/stream-reactor/34?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

hsomyk:
@andrewstevenson actually, jedis version was correct. The issue was with the lib with embedded redis that was used for tests. The version of redis was 2.8.9 and it doesn't support GEOADD commands(available since 3.2.0), so I've changed dependency to a new lib that contains redis with version 4.0.10.

PR #600, opened 2019-07-24
Fix bug in FTPMonitor leading to the buffer being twice as big

jbruggem:
This PR fixes a small bug in the FTPMonitor that lead to slices sizes to be double what they should be.

---
This change is [Reviewable](https://reviewable.io/reviews/landoop/stream-reactor/600)


andrewstevenson:
@jbruggem Thanks!

PR #602, opened 2019-07-25
Update connector to use ftps protocol

GDegrove:
Update `FTPMonitor` to ensure FTPS compatibility by enforcing private mode.

This is made by issuing two cmd in the connection flow of the FTP client. `PBSZ 0` and `PROT P` as metionned in the [original spec of the ftp over tls](https://tools.ietf.org/html/rfc4217)

Based on this [blog post](http://eng.wealthfront.com/2016/06/10/connecting-to-an-ftps-server-with-ssl-session-reuse-in-java-7-and-8/)


GDegrove:
Closes #603

andrewstevenson:
@GDegrove Awesome, thanks!

Issue #559, opened 2019-03-19
MQTT Source wildcard issue: Invalid connect.source.converter.avro.schemas is not configured for topic


caevv:
Version: 1.2.1
Kafka version: 2.1

Hi, I’m getting this error on a MQTT Source
```Caused by: org.apache.zookeeper.server.quorum.QuorumPeerConfig$ConfigException: Invalid connect.source.converter.avro.schemas is not configured for v1.mysite.studio/dev-123/computer/sensor/yoctopuce/gyroscope```

Config is this:
```
"connect.mqtt.kcql": "INSERT INTO spm_record SELECT * FROM v1.mysite.studio/+/computer/sensor/yoctopuce/gyroscope WITHCONVERTER=`com.datamountaineer.streamreactor.connect.converters.source.AvroConverter` WITHKEY(imo)", 
"connect.source.converter.avro.schemas":"v1.mysite.studio/+/computer/sensor/yoctopuce/gyroscope=/avroschema/spm_record.avro"
```

If I don't use wildcard it works.

Full error trace:

> [2019-03-19 15:14:01,593] ERROR Error handling message with id:1 on topic:v1.mysite.studio/dev-123/computer/sensor/yoctopuce/gyroscope (com.datamountaineer.streamreactor.connect.mqtt.source.MqttManager)
org.apache.zookeeper.server.quorum.QuorumPeerConfig$ConfigException: Invalid connect.source.converter.avro.schemas is not configured for v1.mysite.studio/dev-123/computer/sensor/yoctopuce/gyroscope
at com.datamountaineer.streamreactor.connect.converters.source.AvroConverter$$anonfun$1.apply(AvroConverter.scala:50)
at com.datamountaineer.streamreactor.connect.converters.source.AvroConverter$$anonfun$1.apply(AvroConverter.scala:50)
at scala.collection.MapLike$class.getOrElse(MapLike.scala:128)
at scala.collection.AbstractMap.getOrElse(Map.scala:59)
at com.datamountaineer.streamreactor.connect.converters.source.AvroConverter.convert(AvroConverter.scala:50)
at com.datamountaineer.streamreactor.connect.mqtt.source.MqttManager.messageArrived(MqttManager.scala:89)
at org.eclipse.paho.client.mqttv3.internal.CommsCallback.deliverMessage(CommsCallback.java:499)
at org.eclipse.paho.client.mqttv3.internal.CommsCallback.handleMessage(CommsCallback.java:402)
[...]

PR #617, opened 2019-09-05
pass subscribed topic as sourceTopic to avro converter in MQTT connector

ddlenz:
addresses #559

### DIFF
--- a/kafka-connect-mqtt/src/main/scala/com/datamountaineer/streamreactor/connect/mqtt/source/MqttManager.scala
+++ b/kafka-connect-mqtt/src/main/scala/com/datamountaineer/streamreactor/connect/mqtt/source/MqttManager.scala
@@ -100,7 +100,7 @@ class MqttManager(connectionFn: MqttSourceSettings => MqttConnectOptions,
           val scalaList: Seq[String] = l
           scalaList
         }.getOrElse(Seq.empty[String])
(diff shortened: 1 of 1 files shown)


kafka-connect-mqtt/src/main/scala/com/datamountaineer/streamreactor/connect/mqtt/source/MqttManager.scala
```diff
@@ -100,7 +100,7 @@ class MqttManager(connectionFn: MqttSourceSettings => MqttConnectOptions,
           val scalaList: Seq[String] = l
           scalaList
         }.getOrElse(Seq.empty[String])
-        Option(converter.convert(kafkaTopic, topic, message.getId.toString, message.getPayload, keys, kcql.getKeyDelimeter)) match {
+        Option(converter.convert(kafkaTopic, kcql.getSource, message.getId.toString, message.getPayload, keys, kcql.getKeyDelimeter)) match {
```
stheppi: Why would this be better? If non-wildcard subscriptions are used the MQTT topic and the kcql.getSource are the same.

However, on wildcard subscriptions you might not want to provide the expression used ans lose the actual topic name.

ddlenz:
@andrewstevenson this is addressing https://github.com/Landoop/stream-reactor/issues/559 - per the recent comment thread, we have the same issue.
@stheppi we are using wildcard subscriptions per https://github.com/Landoop/stream-reactor/issues/559

andrewstevenson:
closing as stale.

PR #619, opened 2019-09-10
feat(MQTT): replicate shared subscription to all tasks if enabled

c-meier:
Add an option called `connect.mqtt.share.replicate`.

When enabled, the messages published to shared subscriptions are
distributed over all tasks. Therefore, the work generated by a KCQL
instruction with a shared subscription is now divided over all tasks. It
is the job of the MQTT broker to load balance the messages to the
connected tasks.


c-meier:
@andrewstevenson As the MQTT server used in the tests does not support shared subscription, I can't test at that level. However, if you want, I can add some tests to check the distribution of KCQL instructions to the tasks.

andrewstevenson:
@c-meier Lets add a test to check the task split in the connector?

Issue #625, opened 2019-09-25
Support Unix timestamp as double in the InfluxDB Sink connector


afausti:
# Issue Guidelines

Please review these questions before submitting any issue?

### What version of the Stream Reactor are you reporting this issue for?
1.2.1
### Are you running the correct version of Kafka/Confluent for the Stream reactor release?
yes
### Do you have a supported version of the data source/sink .i.e Cassandra 3.0.9?
--
### Have you read the [docs](https://docs.lenses.io/connectors/)?
yes

### What is the expected behaviour?

I would expect the InfluxDB Sink connector to work with Unix timestamps in seconds with type `double`.

### What was observed?

 I've got the following error:

```
java.lang.IllegalArgumentException: Invalid value for field:private_sndStamp.Value '1.3261721783120623E9' is not a valid field for the timestamp
    at com.datamountaineer.streamreactor.connect.influx.writers.TimestampValueCoerce$.apply(TimestampValueCoerce.scala:35)
    at com.datamountaineer.streamreactor.connect.influx.writers.InfluxBatchPointsBuilder$$anonfun$buildPointFromStruct$2$$anonfun$19.apply(InfluxBatchPointsBuilder.scala:237) 
```

I can see that `double` type is not supported [here](https://github.com/Landoop/stream-reactor/blob/master/kafka-connect-influxdb/src/main/scala/com/datamountaineer/streamreactor/connect/influx/writers/TimestampValueCoerce.scala)

afausti:
See PR #626 for a proposal.

andrewstevenson:
Fixed in #626 626

PR #626, opened 2019-09-25
Add support to Unix timestamp as double in the InfluxDB Sink connector

afausti:
This PR is proposal to fix issue #625 

-  It assumes Unix timestamps in seconds if type is double and coerce to Long with milliseconds precision 
- Add test


andrewstevenson:
@afausti Thanks

PR #663, opened 2020-03-28
Add DSE Search Queries (using Solr) to the Cassandra Source Connector

joaomedeiros95:
Adds the ability to run DSE Search Queries on the Cassandra Source Connector.

**Motivation**
Today we're using the stream-reactor Cassandra Source Connector to stream some data from a DSE cluster into some Kafka Topics, with stream-reactor we only have the possibility to make Cassandra queries directly. With DSE we have Solr directly integrated with Cassandra giving us the ability to run Solr queries when the search index is available for the table, in most of the cases Solr queries are faster than Cassandra ones and we don't need to use ALLOW FILTERING on the queries.

This PR adds a new INCREMENTALMODE (dsesearchtimestamp) that will add the possibility to make DSE Search queries on a DSE cluster to be able to speed up those queries when the user has a DSE Cluster.

**What we did**
We just added a new INCREMENTALMODE called **dsesearchtimestamp** that will make a DSE Search queries using Solr instead of a native Cassandra query.

Instead of the native query:
```
SELECT a, b, c, d FROM keyspace.table WHERE pkCol > ? AND pkCol <= ? ALLOW FILTERING;
```

We will have now the query with Solr on the dsesearchtimestamp INCREMENTALMODE:
```
SELECT a, b, c, d FROM keyspace.table WHERE solr_query=?;
```

Where the `solr_query` will be something like this:
```
pkCol:{2020-03-23T15:02:21Z TO 2020-03-23T15:30:12.989Z]
```

**Testing**
You gonna find a new test case on the TestCqlGenerator.

I would probably have implemented two new tests called `TestCassandraSourceTaskDSESearchTimestamp` and `TestCassandraSourceTaskDSESearchTimestampLong` but I didn't find an easy way to do that as the custom index for Solr is not being a requirement inside the project. I would appreciate a little help from you guys on how to implement this.

**Why we don't bind the timestamps**
You're probably asking yourself why we don't bind the timestamps directly on the CQL generated query. That was the idea first but we realized that CQL doesn't allow to bind variables inside a string, so we couldn't do something like that:

```
SELECT a, b, c, d FROM keyspace.table WHERE solr_query='pkCol:{? TO ?]';
```

That's why we mount the Solr query on `CassandraTableReader` and then bind it inside the solr_query.

**Additional comments**
We're already testing this implementation on the staging server and we already processed 15 million DSE rows using this Connect Source implementation.


andrewstevenson:
@joaomedeiros95 Great contribution, thanks!

PR #673, opened 2020-04-21
FTP: respect "connect.ftp.max.poll.records" when reading slices.

### LINKED ISSUE #672: Java Heap OOM Error with kafka-connect-ftp when reading file slice by slice
## Context

* connector: kafka-connect-ftp
* Version: 1.2.1 (reproduced on 1.2.7 too)
* Run with Kafka-cp 4.1.2

## Expected Behavior

The connector can ingest all the files without.

## Observed Behavior

The connector task crashes with a Java Heap Space OutOfMemory  Error 

```
The source emitted an error (it was running since 2020-04-15 15:29:57Z). Trace:
java.lang.OutOfMemoryError: Java heap space
	at java.util.Arrays.copyOf(Arrays.java:3236)
	at java.io.ByteArrayOutputStream.grow(ByteArrayOutputStream.java:118)
	at java.io.ByteArrayOutputStream.ensureCapacity(ByteArrayOutputStream.java:93)
	at java.io.ByteArrayOutputStream.write(ByteArrayOutputStream.java:153)
```

image

## Analysis

Done with @jbruggem 

In the end, the FTP Connector sends the Stream as a List, but [still wrapping the stream so it's still lazy (and uncalculated)](https://github.com/euranova/stream-reactor/blob/master/kafka-connect-ftp/src/main/scala/com/datamountaineer/streamreactor/connect/ftp/source/FtpSourceTask.scala#L129)

Which is then used as a List in KC, which promptly checks its size, which leads to the whole Stream being evaluated.

To prevent this blowing everything up, the original code had a way to only send the top of the Stream and wait for another poll to send the rest.

Unfortunately, when we added the "slice read" mode, [we deactivated that feature with a branching in the code](https://github.com/euranova/stream-reactor/blob/master/kafka-connect-ftp/src/main/scala/com/datamountaineer/streamreactor/connect/ftp/source/FtpSourceTask.scala#L70)

## Fix

Apply the limitation in "slice read" mode too.

image

(PR is coming)

jbruggem:
Closes #672

### DIFF
--- a/kafka-connect-ftp/build.gradle
+++ b/kafka-connect-ftp/build.gradle
@@ -13,21 +13,25 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 project(":kafka-connect-ftp") {
+  ext {
+    pathikritVersion = '2.16.0'
+    ftpserverVersion = '1.0.6'
+    commonsnetVersion = '3.5'
+    commonsCodecVersion = '1.10'
+  }
 
-    ext {
-        pathikritVersion = '2.16.0'
-        ftpserverVersion = '1.0.6'
-        commonsnetVersion = '3.5'
-        commonsCodecVersion = '1.10'
-    }
+  test {
+    // detect memory leaks by constraining the Heap size
+    maxHeapSize = "200m"
+    jvmArgs "-Xms100m"
+  }
 
-    dependencies {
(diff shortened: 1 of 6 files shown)


andrewstevenson:
@jbruggem Thanks!

PR #695, opened 2020-08-10
CLOUD-796 - Introducing an S3 Source for use with Kafka Connect, with…

davidsloan:
CLOUD-796 - Introducing an S3 Source for use with Kafka Connect, with full support for the input formats as outputted by the sink (Avro, Byte, CSV, Json, Parquet, Text).


kafka-connect-aws-s3/README-source.md
```diff
@@ -0,0 +1,183 @@
+# AWS S3 Connector Source Configuration
+
+Before reading this document it is recommended to be familiar with the [general readme for the S3 Connectors](README-sink.md).
+
+The primary use case of the source is to read in files written to S3 by the S3 Sink back into Kafka through Kafka Connect.
```
stheppi, line 5: Not only by S3 sink. maybe there are other producers

kafka-connect-aws-s3/README-source.md
```diff
@@ -0,0 +1,183 @@
...
+
+The options for json, avro and parquet will look like the below:
+
+    connect.s3.kcql=insert into $TOPIC_NAME select * from $BUCKET_NAME:$PREFIX_NAME STOREAS `JSON`
+
+The options for the formats are case-insensitive, but they are presented here in the most readable form.
+
+#### JSON Input Format Configuration
+
+Using JSON as an input format allows you to read in the JSON-written sink files, line by line.
```
stheppi, line 42: let's drop sink files

kafka-connect-aws-s3/README-source.md
```diff
@@ -0,0 +1,183 @@
...
+
+    STOREAS `JSON`
+
+Please note: The JSON is not parsed by the S3 Source connector.  It is recommended to configure the JsonConverter if you wish the Json to be parsed on entry into Kafka.
+
+    value.converter=org.apache.kafka.connect.json.JsonConverter
```
stheppi, line 48: this should use stringconverter otherwise it will wrap the string json in quotes as simple string

stheppi: Json and Text  - there is no difference. unless we support projections in which case yes it makes a difference to parse

kafka-connect-aws-s3/README-source.md
```diff
@@ -0,0 +1,183 @@
...
+
+#### CSV Input Format Configuration
+
+This reads CSV files written to S3 into an Avro struct to be written back to the Kafka source.
+There are 2 options for CSV format, to read CSV files with the column headers or without.
+
+    STOREAS `CSV_WithHeaders`
+    STOREAS `CSV`
+    
+If the headers are included then the Avro message will use these as the field names.
```
stheppi, line 94: We should not advertise CSV (aka text) to AVRO conversion. CSV is the same as Text and Json it is a pass through and should use StringConverter.

kafka-connect-aws-s3/README-source.md
```diff
@@ -0,0 +1,183 @@
...
+    STOREAS `Text`
+    
+It may be required to use the additional configuration options for this connector to ensure that the value is presented as a String.
+
+    value.converter=org.apache.kafka.connect.storage.StringConverter
+    
+
+#### CSV Input Format Configuration
+
+This reads CSV files written to S3 into an Avro struct to be written back to the Kafka source.
```
stheppi, line 88: I don't think we can ever do that unless each field is market as String. Is this happening?

kafka-connect-aws-s3/README-source.md
```diff
@@ -0,0 +1,183 @@
...
+
+
+Using the "With Sizes" options the Source assumes that the files will contain one or two (depending on configuration) 8-byte chunks of data at the start of the file instructing how many bytes to read for the content.
+
+In order to ensure the message is passed through as bytes it may be necessary to set the additional configuration options
+
+    value.converter=org.apache.kafka.connect.converters.ByteArrayConverter
+    key.converter=org.apache.kafka.connect.converters.ByteArrayConverter
+
+Only when using KeyOnly and ValueOnly mode, the size of the chunks of data can be controlled using the additional KCQL option:
```
stheppi, line 123: we should never support this option. No one will be able to re-construct the messages. This works if we read one file and send its content as Value or Key in raw bytes

kafka-connect-aws-s3/README-source.md
```diff
@@ -0,0 +1,183 @@
...
+    BATCH EQUAL batch_size
+    
+For example, if you want a file to be split into ~0.5 MB chunks then you can configure it using this syntax:
+
+    insert into $TOPIC_NAME select * from $BUCKET_NAME:$PREFIX_NAME BATCH=500000 STOREAS `Bytes_ValueOnly`
+
+
+## Partitioning Options
+
+When using the S3 Source it is required to specify partitioning options so that the connector understands the layout of the file system.
```
stheppi, line 134: How does this work if we have CSV ina custom partition with 3 fields?
If there's a custom partitioner it means the fields and values need to be appended.  Do we do that?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3Config.scala
```diff
@@ -108,6 +102,16 @@ object Format extends Enum[Format] {
 
   case object Bytes extends Format
 
+
+  def apply(format: String): Format = {
```
stheppi, line 106: enumeratum already defines a function to read and returnf from string.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/BytesFormatChunkedStreamReader.scala
```diff
@@ -0,0 +1,59 @@
...
+package io.lenses.streamreactor.connect.aws.s3.formats
+
+import java.io.InputStream
+
+import io.lenses.streamreactor.connect.aws.s3.config.BytesWriteMode
+import io.lenses.streamreactor.connect.aws.s3.model.{BucketAndPath, ByteArraySourceData, BytesOutputRow}
+
+import scala.util.Try
+
+class BytesFormatChunkedStreamReader(inputStreamFn: () => InputStream, fileSizeFn: () => Long, bucketAndPath: BucketAndPath, bytesWriteMode: BytesWriteMode, chunkSizeBytes: Int) extends S3FormatStreamReader[ByteArraySourceData] {
```
stheppi, line 26: We cannot support this. If we split binary content in chunks there is no way it can be pieced back correctly

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvFormatStreamReader.scala
```diff
@@ -0,0 +1,81 @@
...
+  override def hasNext: Boolean = iterator.hasNext
+
+  def discoverColumnHeaders(): Option[Array[String]] = {
+
+    if (readHeaders && iterator.hasNext) {
+      lineNumber += 1
+      Some(iterator.next())
+    } else {
+      iterator.peek match {
+        case Some(value) => inventColumnHeaders(value.length)
```
stheppi, line 46: If headers are present, we do not read them - it's just skip.  otherwise it's a case of building a structure using the CSV headers as fields and all the values are string (we cannot infer type optimally).

we should never stream the headers/columns. a client would have no clue how to distinguish a record for headers and one for an entry. and even worst if you have multiple producers to a topic with interleaved records

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvFormatStreamReader.scala
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  override def getLineNumber: Long = lineNumber
+
+  override def close(): Unit = {
+    Try {
+      csvReader.close()
+    }
+    Try {
+      inputStreamReader.close()
```
stheppi, line 74: This should be closed by the previous one

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvFormatStreamReader.scala
```diff
@@ -0,0 +1,81 @@
...
+          throw new IllegalStateException("No rows available from which to create column headers")
+        }
+      }
+
+    }
+
+  }
+
+  override def next(): CsvSourceData = {
+    if (columnHeaders.isEmpty) {
```
stheppi, line 59: CSV passthrough will be hard if the file has headers. For that the user can chose Text (not quite intuitive) which is ok. But if the file has a header and they want to skip it, there's no way for that to happen.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvReaderIteratorAdaptor.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  private var currentRecord: Option[Array[String]] = None
+  private var nextRecord: Option[Array[String]] = Option(csvReader.readNext())
+
+  override def hasNext: Boolean = {
+    nextRecord.nonEmpty
+  }
+
+  def peek: Option[Array[String]] = {
+    hasNext
```
stheppi, line 31: this call is redundant

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvReaderIteratorAdaptor.scala
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package io.lenses.streamreactor.connect.aws.s3.formats
+
+import au.com.bytecode.opencsv.CSVReader
+
+class CsvReaderIteratorAdaptor(csvReader: CSVReader) extends Iterator[Array[String]] {
+
+  private var currentRecord: Option[Array[String]] = None
+  private var nextRecord: Option[Array[String]] = Option(csvReader.readNext())
```
stheppi, line 24: What happens here if there's nothing left to read. Does `csvReader.readNext` throw an exception? if so this code is not handling it.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/FileSizeCounter.scala
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.lenses.streamreactor.connect.aws.s3.formats
+
+class FileSizeCounter(fileSize: Long) {
```
stheppi, line 19: You can use AtomicLong instead of this class.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/ParquetFormatStreamReader.scala
```diff
@@ -0,0 +1,62 @@
...
+
+    lineNumber += 1
+    val nextRec = iterator.next()
+
+    val asConnect = convertToConnectModel(nextRec)
+    SchemaAndValueSourceData(asConnect, lineNumber)
+
+  }
+
+  private def convertToConnectModel(nextRec: GenericRecord): SchemaAndValue = {
```
stheppi, line 58: why not use the body of this method in `next()`?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/ParquetReaderIteratorAdaptor.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  private var currentRecord: Option[GenericRecord] = None
+  private var nextRecord: Option[GenericRecord] = Option(parquetReader.read())
+
+  override def hasNext: Boolean = {
+    nextRecord.nonEmpty
+  }
+
+  def peek: Option[GenericRecord] = {
+    hasNext
```
stheppi, line 32: This call does nothing

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/ParquetReaderIteratorAdaptor.scala
```diff
@@ -0,0 +1,43 @@
...
+    nextRecord
+  }
+
+  override def next(): GenericRecord = {
+
+    currentRecord = nextRecord
+    nextRecord = Option(parquetReader.read())
+
+    currentRecord.getOrElse(throw new IndexOutOfBoundsException("Unable to call next on reaching the end"))
```
stheppi, line 41: Probably we can improve the error message. We can say invalid state reached. The file content has been consumed and there should not be a call for `next` anymore.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/BytesOutputRow.scala
```diff
@@ -0,0 +1,100 @@
...
+
+object BytesOutputRow {
+
+  def apply(storedByteArray: Array[Byte], bytesWriteMode: BytesWriteMode): BytesOutputRow = {
+
+    if (bytesWriteMode == KeyOnly) {
+      new BytesOutputRow(None, None, storedByteArray, Array.empty)
+
+    } else if (bytesWriteMode == ValueOnly) {
+      new BytesOutputRow(None, None, Array.empty, storedByteArray)
```
stheppi, line 34: Why Array.empty? this will send to kafka a 0 bytes array. Do we  make sure we end up creating a SourceRecord with null for the Key when we pass it to Connect?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/BytesOutputRow.scala
```diff
@@ -0,0 +1,100 @@
...
+
+    } else {
+      throw new IllegalArgumentException(s"Invalid apply function for bytesWriteMode key/value only $bytesWriteMode")
+    }
+  }
+
+  def apply(inputStream: InputStream, bytesWriteMode: BytesWriteMode): BytesOutputRow = {
```
stheppi, line 41: It's cleaner to create a reader from byteswritemode which takes the InputStream and avoid the if else. The logic of how it is read is then in the implementation. Easier to test, easier to extend.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/BytesOutputRow.scala
```diff
@@ -0,0 +1,100 @@
...
+
+case class BytesOutputRow(
+                           keySize: Option[Long],
+                           valueSize: Option[Long],
+                           key: Array[Byte],
+                           value: Array[Byte],
+                           bytesRead: Option[Int] = None
+                         ) {
+
+  def toByteArray: Array[Byte] = {
```
stheppi, line 91: What is this for? we merge the key and value into one Array. 
When and why would this be done like this?

davidsloan: This is used in the Sink to build the byte array that is output to the file on S3.

kafka-connect-aws-s3/README-source.md
davidsloan: The opencsv CSV reader provides us with String representations of all of the fields that were written out to CSV.

The converting to struct is an approach suggested by @andrewstevenson so it would be good to get him involved in the discussion.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/ParquetReaderIteratorAdaptor.scala
davidsloan: Good spot, it was used once, but no longer.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/FileSizeCounter.scala
davidsloan: Awesome

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/S3BucketReaderManager.scala
```diff
@@ -0,0 +1,156 @@
...
+
+    var allResults: Vector[PollResults] = Vector()
+
+    var n = 0
+    var allLimit: Int = limit
+    do {
+      chooseReader() match {
+        case Some(currentReader) =>
+          val newResults = resultReader.retrieveResults(currentReader, allLimit)
+          state = if (newResults.isEmpty || (newResults.nonEmpty && newResults.get.resultList.size < allLimit)) {
```
stheppi, line 67: Best is to pattern match on the newResults than doing the forced .get

stheppi: Isn't this potentially taking a long time? 

If I have 1 000 000 files each with 1 000 000 records it could take hours to process them. The while loop terminates when all have been exhausted no?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvReaderIteratorAdaptor.scala
davidsloan: If there's nothing left to read it will return null, which is why it's wrapped in Option()

It does throw an IOException if (and I quote from the jdoc) "bad things happen during the read".

I was of the opinion that if that happens it's going to error out anyway, but perhaps we want to catch it and throw our own exception?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvReaderIteratorAdaptor.scala
davidsloan: how so?  I believe it is used so that it can peek at the next record here:
https://github.com/lensesio/stream-reactor/blob/f2dc893288a8b5c28bf6fd06fca0858aba9ab4c8/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvFormatStreamReader.scala#L45

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/BytesOutputRow.scala
davidsloan: This is used in the sink.

If no key is expected then the key is set to empty.

Alternatively it could be an Option[Array[Byte]] but we already know not to expect a value as we have an Option for the keySize which ultimately ensures this will not be output to the file.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/S3BucketReaderManager.scala
davidsloan: I'd be keen to discuss alternatives.

kafka-connect-aws-s3/README-source.md
davidsloan: Please can we discuss this with @andrewstevenson just to confirm.

csv

kafka-connect-aws-s3/README-source.md
davidsloan: So (just to confirm) take out the chunking mechanism but still enable reading ValueOnly/KeyOnly in the case of single files?  What do we do if the file goes over the Kafka limit?

kafka-connect-aws-s3/README-source.md
davidsloan: The regex largely ignores the partition names and just focuses on the last part of the path.

To construct the objects the configuration is required.

It might be possible to refactor so as to reduce the configuration for the source, but for now this works.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3Config.scala
davidsloan: It's used 3 lines below
Format.withNameInsensitiveOption()

This function serves to add a specific error message and to supply a default when no format has been defined.

kafka-connect-aws-s3/README-source.md
stheppi: We let the runtime fail. We wouldn't know the topic setting for message limit

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvReaderIteratorAdaptor.scala
stheppi: hasNext calls nextRecord.nonEmpty = it does nothing. 
nextRecords is a var not a function.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/BytesOutputRow.scala
```diff
@@ -0,0 +1,100 @@
...
+    val buffer = new ListBuffer[Byte]()
+    keySize.foreach(buffer += _.byteValue())
```
stheppi, line 93: this is not right. you take a Long and convert it to byte and append to the size. You do not append all the Long bytes to the buffer. 

```
val buffer = ByteBuffer.allocate(Long.BYTES)
    buffer.putLong(x)
    buffer.array()
```

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/ParquetReaderIteratorAdaptor.scala
davidsloan: Error message has been changed to:

Invalid state reached: invalid state reached. The file content has been consumed, no further calls to next() are possible.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvReaderIteratorAdaptor.scala
davidsloan: I misunderstood the comment and thought perhaps the method wasn't necessary

I've removed the unnecessary evaluation of hasNext

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvFormatStreamReader.scala
davidsloan: Fixed.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/BytesOutputRow.scala
davidsloan: I have addressed this bug.

We will need to bugfix the sink in master as this code is present there too. (it is only moved in this PR)

I will raise a PR tomorrow.

kafka-connect-aws-s3/README-source.md
davidsloan: CSV parsing has now been removed.

kafka-connect-aws-s3/README-source.md
davidsloan: CSV parsing has now been removed.

kafka-connect-aws-s3/README-source.md
davidsloan: Chunking has now been removed.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/BytesFormatChunkedStreamReader.scala
davidsloan: Chunking has now been removed.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvFormatStreamReader.scala
davidsloan: CSV parsing has now been removed.
Headers are no longer read.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvFormatStreamReader.scala
davidsloan: CSV parsing has now been removed.  The CSV mode still exists with an option to skip the header.  Same configuration as before.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/BytesOutputRow.scala
davidsloan: I have done what you suggest, it has cut down the logic a lot, very nice suggestion.

stheppi (approved):

PR #700, opened 2020-08-20
Add Bucket TimeSeries mode to the Cassandra Source Connector

joaomedeiros95:
We came out with a new incremental mode (different from #663) to avoid ALLOW FILTERING.

**Motivation**
Today we're using the stream-reactor Cassandra Source Connector to stream some data from a Cassandra cluster into some Kafka Topics, with stream-reactor today we only have one option to avoid ALLOW FILTERING on incremental mode, that is using DSESearch.

This PR adds a new INCREMENTALMODE (buckettimeseries) that will add the possibility to make the queries using a technique pretty common on time-series databases, in the case of Cassandra modeling the partition_key is a bucket that divides the documents based on the type of the bucket (MINUTE, HOUR, DAY...) and has a clustering order field that allows the connector to do the **between** operations.

**What we did**
We just added a new INCREMENTALMODE called buckettimeseries that will make queries to Cassandra using the bucket technique.

Instead of the native query:
```sql 
SELECT a, b, c, d FROM keyspace.table WHERE pkCol > ? AND pkCol <= ? ALLOW FILTERING;
```

We will have now the query with Solr on the dsesearchtimestamp INCREMENTALMODE:
```sql
SELECT a, b, c, d FROM keyspace.table WHERE pkCol > ? AND pkCol <= ? AND bucketFieldName IN ?;
```

Where we will have a list of buckets that needs to be used to cover all the time between the partition key query.

**New required configurations**

To this mode work well we need to specify three configurations.

- bucket.timeseries.mode
- bucket.timeseries.format
- bucket.timeseries.field.name

You can find more about them on `CassandraConfigConstants.scala`

**Testing**
You'll gonna find a new test case on the `TestCqlGenerator` and a completely new test file called `CassandraUtilsTest`.

**Additional comments**
We're using this implementation in production for a couple of days and we didn't face any problem, in fact we saw a huge performance improvement from #663.


andrewstevenson (approved):
@joaomedeiros95 Great contribution. Thanks!

PR #701, opened 2020-08-24
Fix/confluent guava versions

andrewstevenson:


andmarios (approved):
Tested Cassandra connector and works.

PR #709, opened 2020-09-30
Fix partition fields lookup

stheppi:
Kafka Connect gets field names in uppercase whilst Hive converts everything to lowercase. When the connector tries to get the value of a partition key (in this case sl_rzbk) Kafka Connect throws an exception because it doesn't know a field with that name


davidsloan:
Would be nice if the correct behaviour were described in a unit test

davidsloan (approved):

andrewstevenson (approved):

PR #726, opened 2021-01-11
Offset storage format

davidsloan:
Change offset storage format to optimise offset read. Add minimal logging for debug purposes.
Fixing offset seek.


davidsloan:
Incorrectly opened.

PR #732, opened 2021-01-14
Feat/azure clouds

andrewstevenson:
Adds 

- Azure Service Bus source and sink
- Azure Queue source and sink
- Azure Table storage sink


michaelandrepearce:
Whats the status on this?

PR #734, opened 2021-01-19
Bump influxdb-java from version 2.9 to 2.21

afausti:
Bump the influxdb-java dependency. In particular, need a fix introduced in influxdb-java 2.16 which skips fields with NaN and Infinity values when writing to InfluxDB.

### DIFF
--- a/kafka-connect-influxdb/build.gradle
+++ b/kafka-connect-influxdb/build.gradle
@@ -17,7 +17,7 @@
 project(":kafka-connect-influxdb") {
 
     ext {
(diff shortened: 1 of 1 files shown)


afausti:
Looks like there's a test failing for the kafka-connect-aws-s3  connector, not related to this PR

afausti:
I have tested this locally with influxdb-java 2.21 and it correctly skips fields with NaN and Infinity values when writing to InfluxDB, so it would be great to have that.

andrewstevenson (approved):

PR #762, opened 2021-04-20
AWS S3 - Nested Data Partitioning (CLOUD-1639)

davidsloan:
Introducing nested data partitioning for AWS S3 Sink.

Support with Structs, Maps, and Arrays in Keys, Values and Headers.


kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/CsvFormatWriter.scala
```diff
@@ -46,8 +45,8 @@ class CsvFormatWriter(outputStreamFn: () => S3OutputStream, writeHeaders: Boolea
     if (!fieldsWritten) {
       writeFields(valueSinkData.schema().orNull)
     }
-    val structValueLookupFn: Option[String] => Option[String] = FieldValueToStringConverter.lookupFieldValueFromSinkData(valueSinkData)
-    val nextRow = fields.map(f => structValueLookupFn(Some(f)) match {
+    val structValueLookupFn: Option[PartitionNamePath] => Option[String] = SinkDataExtractor.extractPathFromSinkData(valueSinkData)
```
stheppi, line 48: This line is a bit hard to read. 

Here is a version for the entire `write` function:
`override def write(keySinkData: Option[SinkData], valueSinkData: SinkData, topic: Topic): Unit = {
    if (!fieldsWritten) {
      writeFields(valueSinkData.schema().orNull)
    }
    val nextRow = fields.map(f => SinkDataExtractor.extractPathFromSinkData(valueSinkData)(Some(PartitionNamePath(f))).orNull)
    csvWriter.writeNext(nextRow: _*)
    csvWriter.flush()
  }`


Also `def extractPathFromSinkData(sinkData: SinkData)(fieldNameOpt: Option[PartitionNamePath]): Option[String]` implementation contract says Option, yet  we throw exception if the field is not specified.

`throw new IllegalArgumentException("FieldName not specified")` 

Is there a reason to have an Option if the field is required?

As a coding style is best to encode the return in an Either[Failure, Result] rather than throwing an exception

davidsloan: > This line is a bit hard to read.

> Here is a version for the entire write function:

Thank you!  I have incorporated it, however it has changed since due to other refactoring.

> Also def extractPathFromSinkData(sinkData: SinkData)(fieldNameOpt: Option[PartitionNamePath]): Option[String] implementation contract says Option, yet we throw exception if the field is not specified.
> throw new IllegalArgumentException("FieldName not specified")
> Is there a reason to have an Option if the field is required?

Yes.  It is not required for Primitives or ByteArrays, however it is required for complex types.  I'd be open to alternative ideas for representing this in the method contract as it is _sometimes_ required.

> As a coding style is best to encode the return in an Either[Failure, Result] rather than throwing an exception

Thank you for this.  I have refactored all of the new Extractors to use this coding style.  However due to interfacing with the rest of the code I have added an adaptor (ExtractorErrorAdaptor) to keep existing behaviour for the time being (so as to not have to carry forward this pattern throughout the rest of the code, due to time constraints).

stheppi (approved):

PR #768, opened 2021-05-06
S3 Connector Error Handling and Retry Logic

davidsloan:
This PR introduces multiple properties for error handling and retry.

``connect.s3.error.policy``
Specifies the action to be taken if an error occurs while inserting the data.
 There are three available options:
    NOOP - the error is swallowed
    THROW - the error is allowed to propagate.
    RETRY - The exception causes the Connect framework to retry the message. The number of retries is set by connect.s3.max.retries.
All errors will be logged automatically, even if the code swallows them.

``connect.s3.max.retries``
The maximum number of times to try the write again.

``connect.s3.retry.interval``
The time in milliseconds between retries.

``connect.s3.http.max.retries``
Number of times to retry the http request, in the case of a resolvable error on the server side.

``connect.s3.http.retry.interval``
If greater than zero, used to determine the delay after which to retry the http request in milliseconds.  Based on an exponential backoff algorithm.

Note: This is a work in progress.

### DIFF
--- a/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3WriterManager.scala
+++ b/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3WriterManager.scala
@@ -18,18 +18,15 @@
 package io.lenses.streamreactor.connect.aws.s3.sink
 
 import com.datamountaineer.streamreactor.common.errors.{ErrorHandler, ErrorPolicy}
-import com.typesafe.scalalogging.{LazyLogging, StrictLogging}
+import com.typesafe.scalalogging.StrictLogging
 import io.lenses.streamreactor.connect.aws.s3.formats.S3FormatWriter
 import io.lenses.streamreactor.connect.aws.s3.model._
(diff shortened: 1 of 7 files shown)


andrewstevenson (approved):

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3WriterManager.scala
```diff
@@ -102,19 +108,28 @@ class S3WriterManager(sinkName: String,
   }
 
   def write(topicPartitionOffset: TopicPartitionOffset, messageDetail: MessageDetail): Unit = {
-    logger.debug(s"[$sinkName] Received call to S3WriterManager.write for ${topicPartitionOffset.topic}-${topicPartitionOffset.partition}:${topicPartitionOffset.offset}")
+    try {
```
stheppi, line 111: `handleTry`  should basically be wrapping the `write` function:


```
 def write(topicPartitionOffset: TopicPartitionOffset, messageDetail: MessageDetail): Unit = {
   handleTry{
     Try{
       <function body>
     }
  }
}
```

So you don't need to put the try-catch yourself

andrewstevenson (approved):

PR #774, opened 2021-05-07
Improve MqttManagerTest stability

johnathana:


johnathana:
> The magic numbers. Is there a way we can sleep and poll the mqtt server (we poll on lower time window)

I am not a fan either of magic numbers and Thread.sleep for testing, but this will require a rewrite of MqttManager to expose internal state.

Issue #737, opened 2021-01-26
kafka-influxdb-sink cannot handle Avro type array anymore


afausti:
# Issue Guidelines

### What version of the Stream Reactor are you reporting this issue for?
2.1.3

### Are you running the correct version of Kafka/Confluent for the Stream reactor release?
Yes, with Confluent Kafka 5.5.2

### Have you read the [docs](https://docs.lenses.io/connectors/)?
Yes

### What is the expected behaviour?

I would expect kafka-influxdb-sink 2.1.3 to support Avro records of type array (see PR #522) In fact I contributed that PR in the past, but I didn't include a test for it  :( 

I can try to fix this with some guidance.

### What was observed?

```
docker-compose exec schema-registry kafka-avro-console-producer --bootstrap-server broker:29092 --topic foo --property value.schema='{"type":"record", "name":"foo", "fields":[{"name":"bar","type":"string"}, {"name":"baz","type":{"type":"array","items":"float"}}]}'

{"bar": "John Doe","baz": [1,2,3]}
```

### What is your Connect cluster configuration (connect-avro-distributed.properties)?
```
$docker-compose exec connect cat ./etc/schema-registry/connect-avro-distributed.properties

bootstrap.servers=localhost:9092
group.id=connect-cluster
key.converter=io.confluent.connect.avro.AvroConverter
key.converter.schema.registry.url=http://localhost:8081
value.converter=io.confluent.connect.avro.AvroConverter
value.converter.schema.registry.url=http://localhost:8081
config.storage.topic=connect-configs
offset.storage.topic=connect-offsets
status.storage.topic=connect-statuses
config.storage.replication.factor=1
offset.storage.replication.factor=1
status.storage.replication.factor=1
internal.key.converter=org.apache.kafka.connect.json.JsonConverter
internal.value.converter=org.apache.kafka.connect.json.JsonConverter
internal.key.converter.schemas.enable=false
internal.value.converter.schemas.enable=false
plugin.path=/usr/share/java,/usr/share/confluent-hub-components
```

### What is your connector properties configuration (my-connector.properties)?

```
docker-compose run kafkaconnect config influxdb-sink
Creating angelofausti_kafkaconnect_run ... done
{
    "connect.influx.db": "mydb",
    "connect.influx.error.policy": "THROW",
    "connect.influx.kcql": "INSERT INTO foo SELECT * FROM foo WITHTIMESTAMP sys_time()",
    "connect.influx.max.retries": "10",
[...]

PR #785, opened 2021-06-05
Added array handling support for influxdb connector

johnathana:


andrewstevenson (approved):

afausti:
That's great! thank you @stheppi and @johnathana for implementing this. I think it closes #737

PR #787, opened 2021-06-10
Local Write Mode - Build Locally, then Ship

davidsloan:
Enabling the ability to build the file locally before uploading in one go in the commit.

This enables new configuration options to ensure the files are locally built and complete before uploading to S3.  This avoids the necessity for a rename and therefore doesn't clutter version history.

The new configuration options are:
`connect.s3.write.mode` options are `BuildLocal` (which builds the file locally before uploading) or `Streamed` (which keeps the previous behaviour of streaming)
`connect.s3.local.tmp.directory` is optional to supply a directory to write the files to locally.  If none is supplied and `BuildLocal` mode is used, then a directory will be created in your system temporary directory (eg /tmp)

The new `BuildLocal` write mode is currently limited to 5GB as it does not use the multipart upload API.

### DIFF
--- a/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala
+++ b/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala
@@ -17,41 +17,16 @@
 package io.lenses.streamreactor.connect.aws.s3.config
 
 import com.datamountaineer.streamreactor.common.config.base.traits.BaseSettings
-import io.lenses.streamreactor.connect.aws.s3.config.S3ConfigSettings.{LOCAL_TMP_DIRECTORY, WRITE_MODE}
-import io.lenses.streamreactor.connect.aws.s3.config.S3WriteModeSettings.defaultWriteMode
-import io.lenses.streamreactor.connect.aws.s3.model.{LocalLocation, S3WriteMode}
-import io.lenses.streamreactor.connect.aws.s3.model.S3WriteMode.{BuildLocal, Streamed}
-
-import java.nio.file.Files
-import java.util.UUID
-
-object S3WriteModeSettings {
-
-  val defaultWriteMode: S3WriteMode = Streamed
-
-}
+import io.lenses.streamreactor.connect.aws.s3.config.S3ConfigSettings.WRITE_MODE
+import io.lenses.streamreactor.connect.aws.s3.model.S3OutputStreamOptions
 
 trait S3WriteModeSettings extends BaseSettings {
 
-  def s3WriteMode() : S3WriteMode = {
-    S3WriteMode
-      .withNameInsensitiveOption(getString(WRITE_MODE))
-      .getOrElse(defaultWriteMode)
-  }
-
-  def s3LocalBuildDirectory(sinkName: Option[String]) : Option[LocalLocation] = {
-    s3WriteMode() match {
-      case S3WriteMode.Streamed => Option.empty[LocalLocation]
-      case S3WriteMode.BuildLocal => Option(getString(LOCAL_TMP_DIRECTORY))
-        .filter(_.trim.nonEmpty)
-        .orElse(createTmpDir(sinkName))
-        .map(LocalLocation(_))
+  def s3WriteOptions(props: Map[String,String]) : S3OutputStreamOptions = {
+    S3OutputStreamOptions(getString(WRITE_MODE), props) match {
+      case Left(exception) => throw exception
+      case Right(value) => value
     }
   }
 
-  private def createTmpDir(sinkName: Option[String]): Option[String] = {
-    val sinkIdent = sinkName.getOrElse("MissingSinkName")
-    val uuid = UUID.randomUUID().toString
-    Some(Files.createTempDirectory(s"$sinkIdent.$uuid").toAbsolutePath.toString)
-  }
 }
--- a/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala
+++ b/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala
@@ -22,11 +22,8 @@ import io.lenses.streamreactor.connect.aws.s3.model.S3OutputStreamOptions
 
 trait S3WriteModeSettings extends BaseSettings {
 
-  def s3WriteOptions(props: Map[String,String]) : S3OutputStreamOptions = {
-    S3OutputStreamOptions(getString(WRITE_MODE), props) match {
-      case Left(exception) => throw exception
-      case Right(value) => value
-    }
+  def s3WriteOptions(props: Map[String,String]) : Either[Exception,S3OutputStreamOptions] = {
+    S3OutputStreamOptions(getString(WRITE_MODE), props)
   }
 
 }
--- a/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala
+++ b/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2021 Lenses.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.lenses.streamreactor.connect.aws.s3.config
+
+import com.datamountaineer.streamreactor.common.config.base.traits.BaseSettings
+import io.lenses.streamreactor.connect.aws.s3.config.S3ConfigSettings.{LOCAL_TMP_DIRECTORY, WRITE_MODE}
+import io.lenses.streamreactor.connect.aws.s3.config.S3WriteModeSettings.defaultWriteMode
+import io.lenses.streamreactor.connect.aws.s3.model.{LocalLocation, S3WriteMode}
+import io.lenses.streamreactor.connect.aws.s3.model.S3WriteMode.{BuildLocal, Streamed}
+
+import java.nio.file.Files
+import java.util.UUID
+
+object S3WriteModeSettings {
+
+  val defaultWriteMode: S3WriteMode = Streamed
+
+}
+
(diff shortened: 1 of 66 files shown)


kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala
```diff
@@ -0,0 +1,57 @@
...
+      case S3WriteMode.Streamed => Option.empty[LocalLocation]
+      case S3WriteMode.BuildLocal => Option(getString(LOCAL_TMP_DIRECTORY))
+        .filter(_.trim.nonEmpty)
+        .orElse(createTmpDir(sinkName))
+        .map(LocalLocation(_))
+    }
+  }
+
+  private def createTmpDir(sinkName: Option[String]): Option[String] = {
+    val sinkIdent = sinkName.getOrElse("MissingSinkName")
```
stheppi, line 53: if sinkName is None it might allow to sinks to share the same name and there will be problems. It's best to enforce the sinkName to be String

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala
```diff
@@ -0,0 +1,57 @@
...
+
+trait S3WriteModeSettings extends BaseSettings {
+
+  def s3WriteMode() : S3WriteMode = {
+    S3WriteMode
+      .withNameInsensitiveOption(getString(WRITE_MODE))
+      .getOrElse(defaultWriteMode)
+  }
+
+  def s3LocalBuildDirectory(sinkName: Option[String]) : Option[LocalLocation] = {
```
stheppi, line 42: if it's asking for local there should not be checking the mode. the code calling this should make sure it builds the details.
Might be good to have a builder which does nothing when streamed is used.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/AvroFormatWriter.scala
```diff
@@ -49,12 +47,10 @@ class AvroFormatWriter(outputStreamFn: () => S3OutputStream) extends S3FormatWri
 
   }
 
-  override def close(): Unit = {
-    avroWriterState.fold(logger.debug("Requesting close when there's nothing to close"))(_.close())
+  override def close(newName: BucketAndPath): Unit = {
```
stheppi, line 50: Is there a reason why this has to be passed to the close function? Wouldn't this information be part of the writer internal state ?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/BucketPathPrefix.scala
```diff
@@ -41,10 +46,36 @@ case class BucketAndPrefix(
     .foreach(_ => throw new IllegalArgumentException("Nested prefix not currently supported"))
 }
 
-case class BucketAndPath(
+trait Location {
```
stheppi, line 49: Is there a reason for being empty and not have `def path:String`?
Can we express the stream and local "location" in a similar way? It looks like an if stream then `nothing really here` else if local folder path

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/S3WriteMode.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import scala.collection.immutable
+
+sealed trait S3WriteMode extends EnumEntry
+
+object S3WriteMode extends Enum[S3WriteMode] {
+
+  override val values: immutable.IndexedSeq[S3WriteMode] = findValues
+
+  case object Streamed extends S3WriteMode
```
stheppi, line 29: These two should be able to create the location and have

```
sealed trait Location {
  def path:String
}

case class Local (path:java.nio.Path) extends Location
case class Remote(path: String) extends Location 
```

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3WriterManager.scala
```diff
@@ -215,12 +221,22 @@ object S3WriterManager {
...
       config.bucketOptions.find(_.sourceTopic == topicPartition.topic.value) match {
         case Some(bucketOptions) =>
-          val fileNamingStrategy = fileNamingStrategyFn(topicPartition.topic)
-
-          val path: BucketAndPath = fileNamingStrategy
-            .stagingFilename(bucketOptions.bucketAndPrefix, topicPartition, partitionValues)
-          val size: Int = minAllowedMultipartSizeFn()
-          S3FormatWriter(bucketOptions.formatSelection, outputStreamFn(path, size))
+          bucketOptions.writeMode match {
+            case S3WriteMode.Streamed => val fileNamingStrategy = fileNamingStrategyFn(topicPartition.topic)
```
stheppi, line 225: See previous comment. The write mode can also be responsible for handling this code. it would make it easier and more composable. Here you would have the writeMode and from there just call .

You have this code:

```
val path: BucketAndPath = fileNamingStrategy.stagingFilename(bucketOptions.bucketAndPrefix, topicPartition, partitionValues)
``` 
duplicated. 

Then 
```bucketOptions.localBuildDirectory``` should never be Option. If you have a Local write mode this should be set. This will simplify the code and won't require checks on an invalid state because wouldn't have been created in the first place.

File: kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala (hunk 1 above)
Comment (davidsloan) [reply kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala:53]: OK.  I will change this so that either `connect.s3.local.tmp.directory` or `name` must be used when attempting to use `BuildLocal` write mode.

File: kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala (hunk 2 above)
Comment (davidsloan) [reply kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala:42]: Good plan.  The configuration is not sensible.  I will refactor the configuration so it contains a top level case class with the relevant values to the selected mode only.

File: kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/AvroFormatWriter.scala (hunk 3 above)
Comment (davidsloan) [reply kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/AvroFormatWriter.scala:50]: The term "writer" applies in the S3 Sink to multiple writers.

We have the S3 Writer and the Format Writer.
The S3 writer has the internal state including topics/partitions etc.
The format writer knows nothing of the internal state and is just handling streams.

As we need to give it a new file name on close then it seems to fit in best here.  It means I was able to get rid of the logic between layers where I was keeping track of whether we have an outstanding rename in the S3 writer after coming to the end of the stream at the Format writer.

It seems to simplify things, anyway.

File: kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/S3WriteMode.scala (hunk 5 above)
Comment (davidsloan) [reply kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/S3WriteMode.scala:29]: This wouldn't work as overriding `def path:String` with `path:java.nio.Path`, however I've followed the majority of the suggestion.

File: kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/BucketPathPrefix.scala (hunk 4 above)
Comment (davidsloan) [reply kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/BucketPathPrefix.scala:49]: There are 3 different types of location and we don't necessarily use `path` in each one.

But I've added it to some extent.

File: kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3WriterManager.scala (hunk 6 above)
Comment (davidsloan) [reply kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3WriterManager.scala:225]: I've made a bunch of changes around the configuration so that it is much improved and more straightforward.

File: kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala [hunk 7]
```diff
@@ -0,0 +1,32 @@
...
+
+package io.lenses.streamreactor.connect.aws.s3.config
+
+import com.datamountaineer.streamreactor.common.config.base.traits.BaseSettings
+import io.lenses.streamreactor.connect.aws.s3.config.S3ConfigSettings.WRITE_MODE
+import io.lenses.streamreactor.connect.aws.s3.model.S3OutputStreamOptions
+
+trait S3WriteModeSettings extends BaseSettings {
+
+  def s3WriteOptions(props: Map[String,String]) : S3OutputStreamOptions = {
```
stheppi, line 25: A small one.
Avoiding coding with throw (Java style) and have the method return Either[E, R] is much better performance wise, code wise (because the caller will know how to react to errors vs catching throwable if we don't document the exceptions), and it's more scala-esque

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/BuildLocalOutputStream.scala
```diff
@@ -0,0 +1,69 @@
...
+import com.typesafe.scalalogging.LazyLogging
+import io.lenses.streamreactor.connect.aws.s3.model.{RemotePathLocation, LocalLocation}
+import org.jclouds.blobstore.domain.{MultipartPart, MultipartUpload}
+
+import java.io.{BufferedOutputStream, File, FileOutputStream, OutputStream}
+import java.nio.ByteBuffer
+
+
+class BuildLocalOutputStream(
+                                      initialName: LocalLocation,
```
stheppi, line 29: the format here looks a bit weird.

stheppi (approved):

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/BuildLocalOutputStream.scala
davidsloan: Fixed.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3WriteModeSettings.scala
davidsloan: Fixed.  Please take a look.

Although I had to throw the error eventually as it must interface with the Java connect API.

So the exception is now being thrown in S3SinkTask.start

I do not think there an alternative to this?

PR #815, opened 2021-11-22
Optimal Sink Seek

davidsloan:
Support for optimal sink offset seek algorithm.

This includes some new configuration:

## Seek Configuration

This version of the connector changes to a new offset seek algorithm that performs the offset seek efficiently.  It does this by keeping new index files within the S3 bucket.

### connect.s3.seek.max.files

The default is _5_.

This specifies the maximum number of index files to keep per topic/partition before raising an error.

If the index files keep accumulating then it means there is an error in the cleanup so the sink cannot continue.

### connect.s3.seek.migration.enabled

The default is _false_.

If you are upgrading a connector from a previous version of the sink, then you will need to specify this property to enable migration.

This option is intended for one-time upgrade usage and will be deprecated in future releases of the connector.

On the first run of any connector then the old slow offset seek will be performed, however the index files will be written so performance will recover after the first offset seek.

### Writing - At Commit Time

1. An index file is written to `.indexes/connectorName/topic/partition/offset`, containing only the path of the target file that will shortly be written.
2. The target file is written.
3. Clean up  - retrieve all files for the given partition, if there are more than one delete all but the last.

### Reading - At Assignment

1. Retrieve all index files for partition and the file contents.
2. Check for the existence of the files referenced in the index files.
3. Delete any that point to non-existent files.
4. Latest offset will be the file remaining.
5. If no file remains, and the optional flag `connect.s3.seek.migration.enabled` is set, an upgrade will be performed:
    - falls back to the previous (inefficient) algorithm, but ensure clear logging.
    - once the correct offset has been calculated, write it to the path.  Therefore if the sink is restarted it will not have to reperform the expensive operation to calculate the offset again.

The `connect.s3.seek.migration.enabled` flag will be removed in a future version.


kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/S3SinkConfig.scala
```diff
@@ -21,32 +21,39 @@ import cats.syntax.all._
...
-  def apply(s3ConfigDefBuilder: S3ConfigDefBuilder): Either[Exception, S3SinkConfig] = {
-    SinkBucketOptions(s3ConfigDefBuilder) match {
-      case Left(ex) => ex.asLeft[S3SinkConfig]
-      case Right(value) => S3SinkConfig(S3Config(s3ConfigDefBuilder.getParsedValues), value).asRight
-    }
+  def apply(s3ConfigDefBuilder: S3ConfigDefBuilder): Either[Throwable, S3SinkConfig] = {
+    for {
+      sinkBucketOptions <- SinkBucketOptions(s3ConfigDefBuilder)
+      offsetSeekerOptions = OffsetSeekerOptions(
+          Try(s3ConfigDefBuilder.getInt(SEEK_MAX_INDEX_FILES).toInt).getOrElse(SEEK_MAX_INDEX_FILES_DEFAULT),
```
stheppi, line 39: if we define a field as boolean/int the Connect API ensures the type so we don't need, technically, to re-validate. Connect will throw an error  if the type is not correct

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
```diff
@@ -0,0 +1,184 @@
...
+  /**
+   * deletes all index files except for the one corresponding to topicPartitionOffset
+   *
+   * @param mostRecentIndexFile the latest offset successfully written
+   * @param topicPartition the topicPartition
+   */
+  def clean(mostRecentIndexFile: RemoteS3PathLocation, topicPartition: TopicPartition): Either[SinkError, Int] = {
+    val remoteS3PathLocation = mostRecentIndexFile.root().withPath(IndexFilenames.indexForTopicPartition(sinkName, topicPartition.topic.value, topicPartition.partition))
+    storageInterface.list(remoteS3PathLocation)
+      .leftMap(e => {
```
stheppi, line 21: you can have 
```
{e =>
}
```

and avoid 
```
(e=>{})
```

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
```diff
@@ -0,0 +1,184 @@
...
+        val logLine = s"Couldn't retrieve listing for (${mostRecentIndexFile.path}})"
+        logger.error("[{}] {}", sinkName, logLine, e.exception)
+        new NonFatalS3SinkError(logLine, e.exception)
+      })
+      .flatMap {
+        indexes: List[String] =>
+          val filtered = indexes.filterNot(_ == mostRecentIndexFile.path)
+          if (indexes.size > maxIndexes) {
+            logAndReturnMaxExceededError(topicPartition, indexes)
+          } else if (filtered.size == indexes.size) {
```
stheppi, line 31: will it ever get into this condition? if we have already a "index" file, will be deleted by the time this method executes?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
```diff
@@ -0,0 +1,184 @@
...
+            val logLine = s"Latest file not found in index (${mostRecentIndexFile.path})"
+            logger.error("[{}] {}", sinkName, logLine)
+            NonFatalS3SinkError(logLine).asLeft
+          } else {
+            storageInterface
+              .deleteFiles(mostRecentIndexFile.bucket, filtered)
+              .map(_ => {
```
stheppi, line 38: same as earlier about
```
.mape{ _ =>

}
```

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
```diff
@@ -0,0 +1,184 @@
...
+   * @return either a SinkError or an option to a TopicPartitionOffset with the seek result.
+   */
+  def seek(topicPartition: TopicPartition, fileNamingStrategy: S3FileNamingStrategy, bucketAndPrefix: RemoteS3RootLocation): Either[SinkError, Option[TopicPartitionOffset]] = {
+
+    /**
+     * Parses the filename of the index file, converting it to a TopicPartitionOffset
+     * @param maybeIndex option of the index filename
+     * @return either an error, or a TopicPartitionOffset
+     */
+    def indexToOffset(maybeIndex: Option[String]): Either[Throwable, Option[TopicPartitionOffset]] = {
```
stheppi, line 106: Is there a scenario when we have None for maybeIndex and the return is None?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
```diff
@@ -0,0 +1,184 @@
...
+      }
+    }
+
+    storageInterface.list(bucketAndPrefix.withPath(
+      IndexFilenames.indexForTopicPartition(sinkName, topicPartition.topic.value, topicPartition.partition)
+    ))
+      .leftMap(e => new NonFatalS3SinkError("Couldn't retrieve listing", e.exception))
+      .flatMap(
+        indexes =>
+          if (indexes.size > maxIndexes) {
```
stheppi, line 122: if we do this, we would never ever recover without someone manually removing the files right. 
So imagine the connector stops. the first reaction is for the user might be to restart. As it starts, this kicks in and fails.

Should we try to clean first before throwing? what. do you think?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
```diff
@@ -0,0 +1,184 @@
...
+            logAndReturnMaxExceededError(topicPartition, indexes)
+          } else if (indexes.isEmpty && fallbackSeeker.nonEmpty) {
+            fallbackOffsetSeek(fallbackSeeker.get, topicPartition, fileNamingStrategy, bucketAndPrefix)
```
stheppi, line 125: if we don't have a migration, shouldn't then be no fallback?
Imagine the scenario we start a connector with the new behaviour and this happens:

1.  it starts writing records
2. after a while it writes the index file with the content pointing to the target file
3. and writes the target file

So will there ever be, apart from migration, a case where we have empty indexes but have files?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
```diff
@@ -0,0 +1,184 @@
...
+   * Given a bucket and a list of files, attempts to load them to establish the most recent valid index
+   *
+   * @param bucketAndPrefix Remote S3 Root Location of the sink files
+   * @param indexFiles      List of index files
+   * @return either a FileLoadError or an optional string containing the valid index file of the offset
+   */
+  def scanIndexes(bucketAndPrefix: RemoteS3RootLocation, indexFiles: List[String]): Either[FileLoadError, Option[String]] = {
+    var latest: Either[FileLoadError, Option[String]] = None.asRight
+
+    indexFiles
```
stheppi, line 165: would it be better to translate to long the index file name and then sort on that?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
```diff
@@ -0,0 +1,184 @@
...
+              targetBucketAndPath = bucketAndPrefix.withPath(targetFileName)
+              pathExists <- storageInterface.pathExists(targetBucketAndPath)
+            } yield {
+              if (pathExists) Some(idxFileName) else Option.empty[String]
+            }
+          }
+        latest.isRight && latest.right.exists(_.isEmpty)
+      }
+
+      latest
```
stheppi, line 181: var field :)
Let's make it a bit more functional. 
I suggest we list all the indexes and split the return into:
1. files with a target file which exists
2. files with a target file which does not exist (this would need to be delete somewhere else)

the point1 list can then be iterated over by the caller and delete anything but last. and this we can do by ordering in descending order and delete the `.tail` of the list

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
```diff
@@ -0,0 +1,184 @@
...
+  /**
+   * Seeks the filesystem to find the latyest offsets for a topic/partition.
+   * @param topicPartition the TopicPartition for which to retrieve the offsets
+   * @param fileNamingStrategy the S3FileNamingStrategy to use in the case that a fallback offset seeker is required.
+   * @param bucketAndPrefix the configured RemoteS3RootLocation
+   * @return either a SinkError or an option to a TopicPartitionOffset with the seek result.
+   */
+  def seek(topicPartition: TopicPartition, fileNamingStrategy: S3FileNamingStrategy, bucketAndPrefix: RemoteS3RootLocation): Either[SinkError, Option[TopicPartitionOffset]] = {
+
+    /**
```
stheppi, line 101: why is this indentation different from above

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
stheppi: It's best to break in smaller functions as well.
The function which takes a the files and reads the content should be one function. 

The result of this function (whatever the shape, it needs to have a index path and Option[target path]) will be the input to another function which can identify the files which need to be deleted and also another function which can identify the latest (let's not optimise for performance - iterating over a few items list it will be fast and there's a lot of I/O in the connector to worry about performance rather than CPU cycles and a bit of extra memory allocation)

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
davidsloan: Good point.  I was thinking this would require manual intervention, for example fixing permissions on AWS.  However once that manual intervention has been performed there's not any reason to redo the cleanup.  I will fix this.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/S3SinkConfig.scala
davidsloan: ok, so I will simplify this to 
`offsetSeekerOptions = OffsetSeekerOptions(
        s3ConfigDefBuilder.getInt(SEEK_MAX_INDEX_FILES),
        s3ConfigDefBuilder.getBoolean(SEEK_MIGRATION)
      )`

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
davidsloan: It's a nested function.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
davidsloan: Yes.  For a brand new connector no files will exist; it will not find an index file, so this will simply return None.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
davidsloan: I think this is marginally more efficient.

With AWS S3 there is a guarantee that [list results are always returned in UTF-8 binary order](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ListingKeysUsingAPIs.html).

As long as we are sure the index files are correctly padded we can rely on the order they are returned from S3.

Which means fewer in-memory sort operations.

For the number of results we're talking about it wouldn't make much difference but I'm thinking it seems to be a good practice.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
davidsloan: My hope is that it never gets into this condition, it is a precondition/safeguard only, this error message would be invaluable for troubleshooting.

Regarding the deleting, the current index will not be deleted until the next commit, so this will be safe, no?  Do you have concerns?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
davidsloan: My thoughts are... if it is not a migration, we don't need a fallback.  The only time the index will be empty is a new connector, if this seeks then we want to return None.

The fallbackSeeker is only set for a migration scenario.  Otherwise it will be empty, and the next condition will be executed.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
davidsloan: > var field :) Let's make it a bit more functional. I suggest we list all the indexes and split the return into:
> 
> 1. files with a target file which exists
> 2. files with a target file which does not exist (this would need to be delete somewhere else)
> 
> the point1 list can then be iterated over by the caller and delete anything but last. and this we can do by ordering in descending order and delete the `.tail` of the list

1. I feel bad/guilty for using the var.  I know there must be some functional alternative
2. I considered similar to this algorithm (above) but the problem with that, given 100 index files (say 0…100) then you are:
    - loading 100 index files to retrieve the source
    - attempting to load 100 target files
Whereas it is possible you only need to load a single file (/100), if this is valid then you can just move on and delete all the rest without 99 * IO to S3 which is over a network and relatively slow.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
stheppi: ah, sorry about this.. maybe i need to look better/more and get new glasses. :)

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/seek/IndexManager.scala
stheppi: Please clarify my understanding:

1. we have the file to upload and we know the last offset
2. we write the index file (the name is the offset value - we do it per partition ). the file contains the target file path we are about to upload.
3. we upload the file. if successful then the index file has a valid reference. otherwise the index file is "orphaned".

When we extract the index
* during topic-partition re-assignment
* during a write (we want to avoid writing orphaned files)

Therefore listing all indexes for the topic-partition should have at most 2 files, in most scenarios(success) it's one. We have two files if we  reached step 2 but 3 failed.

So we list the files and read their content to see where they point two and if the referenced file exists. if the reference file does not exist, the index file:
1. is not included it when calculating the latest
2. needs to be deleted


So if the above understanding is correct, then we should not rely on only picking the latest index file and iterate over all available. Like i said we should not have more than 2 if we do the cleanup. Thoughts?

davidsloan: This is absolutely correct.

However... if there is a problem with file deletion, we do not stop the sink of throw an error, and assume this will be cleaned up later.

An example of such a problem is that the S3 permissions are set to allow the creation of files but not deletion.  Therefore delete operations will fail but we continue on regardless.

Whilst it is an unusual case, is it not worth ensuring the code is still efficient?

stheppi (approved):

iosifnicolae2:
@davidsloan We're seeing the below warning in console and then the task does not sink any files to S3:
```
WARN WorkerSinkTask{id=REDACTED-s3-0} Commit of offsets timed out (org.apache.kafka.connect.runtime.WorkerSinkTask) [task-thread-REDACTED-0]
```

Do you have any idea on how can we solve this problem?

Thank you!

PR #853, opened 2022-05-11
Add aws sts dependency.

caseycrites:
This is required in order for the AWS client to assume roles using a web identity token, like you might have for a pod with an IAM role linked service account in an EKS cluster.

This is in the same vein as [this PR](https://github.com/lensesio/secret-provider/pull/25) against the `secret-provider` repository.


davidsloan (approved):
Thank you!

PR #808, opened 2021-10-13
Extend sink jms connector to support converter classes from runtime and add protobuf sink converter

meghanayofficial:
This PR is responsible for extending below features:

1. The extension of the converters from the configuration properties and also via kcql using WITHFORMAT and WITHCONVERTER

2. Add Protobuf Messaging converter for sink connector. 

3. Passing Headers via record

4. Supporting subscription name for the source connector

PR #855, opened 2022-05-13
Refactor/extend sink proto converter

davidsloan:
Following the upgrade of the Stream Reactor project to SBT, this is a rebase and refactor of #808 .  This aims to simplify the code of the original contributor, making it less procedural and hopefully more functional, tidy up the solution, remove unlicenced third party code, and also to address the declined code review comments.

Original Notes : (from #808):

> 
> This PR is responsible for extending below features:
> 
> The extension of the converters from the configuration properties and also via kcql using WITHFORMAT and WITHCONVERTER
> 
> Add Protobuf Messaging converter for sink connector.
> 
> Passing Headers via record
> 
> Supporting subscription name for the source connector
>


stheppi (approved):
Nice work.

PR #857, opened 2022-05-19
Enforce format

davidsloan:


.github/workflows/build.yml
```diff
@@ -41,6 +41,29 @@ jobs:
           find $HOME/.ivy2/cache                       -name "ivydata-*.properties" -delete || true
           find $HOME/.sbt                              -name "*.lock"               -delete || true
 
+  scalafmt:
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    outputs:
```
davebassan, line 47: I think you can lose the outputs section from this job.

davidsloan: Good point, done!

stheppi (approved):

PR #801, opened 2021-09-22
Sftp connector

politrons:
feat(ftp): Sftp client for Secure File Transfer Protocol
We have the need to use this protocol for a SFTP server we have, and the current protocol solutions that the FTP Connector provide (FTP, FTPS) cannot cover that need.

This feature provides this capacity using JSch (Java Secure Channel)

PR #862, opened 2022-06-17
Support slice mode for SFTP

jbruggem:
When support for [SFTP](https://github.com/lensesio/stream-reactor/pull/801) was added to the FTP connector, it did not include support for the "slice mode", where files are not being read in one pass, but by slices. This mode makes it possible to read huge files without crashing the connector.

This PR adds support for this "slice mode" by adding two missing overrides to `SFTPClient`: `retrieveFileStream` and `setRestartOffset`. 

It also:

* Extracts SFTP tests to a separate module for readability. They were added to `BySlicesTest` but where not actually testing the "slice mode".
* Adds an SFTP test where the slice mode is activated, to test the change introduced in this PR.
* Re-throw exceptions coming from the FTP backend in `FTPMonitor.bySlicesFetch` to expose the cause of failures when a connector fails.


davidsloan:
It looks ok to me.  @politrons would you also like to take a look too before we merge?

davidsloan (approved):

politrons:
LGTM, this is a great contribution, thank you very much.

Issue #874, opened 2022-07-19
AWS S3 source connect surrounds data with quotes


aidan-melen:
# Issue Guidelines

How can I configure the S3 Sink to write data to S3 objects without surrounding quotes? I would expect the `source_topic` and `target_topic` to be identical with a s3 sink and s3 source connects with 0 consumption lag.

following this [aws blog post](https://aws.amazon.com/blogs/big-data/back-up-and-restore-kafka-topic-data-using-amazon-msk-connect/), more or less.

### What version of the Stream Reactor are you reporting this issue for?

I built a fat jar from master branch. The 3.0.1 release won't work because my bucket is in us-west-2 and I found that only the latest build supports the `connect.s3.aws.client=AWS` property.

### Are you running the correct version of Kafka/Confluent for the Stream reactor release?

yes

### Have you read the [docs](https://docs.lenses.io/connectors/)?

yes

### What is the expected behaviour?
The S3 source connector loads objects messages from S3 to Kafka without surrounding quotes.

### What was observed?
![Screen Shot 2022-07-18 at 9 45 15 PM](https://user-images.githubusercontent.com/54326778/179660317-20b9eb6f-e2c1-4a89-b497-1c98244f0cd3.png)

### What is your Connect cluster configuration (s3-sink-connector.properties)?

```ini
connector.class=io.lenses.streamreactor.connect.aws.s3.sink.S3SinkConnector
key.converter.schemas.enable=false
connect.s3.kcql=INSERT INTO mkc-tutorial-destination-bucket:tutorial SELECT * FROM source_topic WITH_FLUSH_INTERVAL = 60 WITH_FLUSH_COUNT = 100
tasks.max=2
topics=source_topic
connect.s3.aws.client=AWS
connect.s3.aws.region=us-west-2
schema.enable=false
value.converter=org.apache.kafka.connect.storage.StringConverter
errors.log.enable=true
key.converter=org.apache.kafka.connect.storage.StringConverter
```

This is what the bucket objects contain:

```
"asdlkjfasdfkjlfsda"
"jhasdfhiughsdafouih"
"igsdfiuhiudyfs"
"ajshdfoifasdoij"
"ijfhiudhhufdshu"
"oihsdfasdfasdf"
"gfahfahafhaf"
"asdfreatsafh"
""
"asdtadsreatreg"
"a"
"aa"
"aaa"
"aaa"
"aaaa"
"aaaaa"
"aaaaa"
"a"
"a"
"aasdasda"
[...]

Issue #869, opened 2022-07-08
AWS S3 Sink connector does not store Kafka message key


YordanPavlov:
### What version of the Stream Reactor are you reporting this issue for?

Latest master build

### What is the expected behaviour?

The S3 sink connector to store the Kafka message along with its key inside S3.

### What was observed?

The Kafka message key is ignored and only the value is stored.

### What is your Connect cluster configuration (connect-avro-distributed.properties)?

```
bootstrap.servers=kafka:9092
key.converter=org.apache.kafka.connect.json.JsonConverter
value.converter=org.apache.kafka.connect.json.JsonConverter
key.converter.schemas.enable=false
value.converter.schemas.enable=false
offset.storage.file.filename=/tmp/connect.offsets
offset.flush.interval.ms=10000
plugin.path=/home/src/kafka/jars
group.id=kafka-connect
storage.topic=_connect-configs
offset.storage.topic=_connect-offsets
status.storage.topic=_connect-status
```

### What is your connector properties configuration (my-connector.properties)?

```
name=S3SinkConnectorS3 # this can be anything
connector.class=io.lenses.streamreactor.connect.aws.s3.sink.S3SinkConnector
tasks.max=1
aws.auth.mode=Default
topics=test_yordan_kafka_connect
connect.s3.kcql="insert into `test:test-bucket` select * from test_yordan_kafka_connect STOREAS `JSON` WITH_FLUSH_COUNT = 5000"
connect.s3.aws.client=AWS
connect.s3.aws.region=eu-central-1
timezone=UTC
errors.log.enable=true
```

I did some debugging of the codebase and suspect the issue is the key is being ignored here:
https://github.com/lensesio/stream-reactor/blob/master/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/formats/JsonFormatWriter.scala#L41

davidsloan:
Hello, thanks for raising this.

Exactly what you ask is not supported because Kafka Connect is flexible and there are other ways to solve this problem.  Just some ideas:

1. You could use SMTs before reaching the connector to insert the keys into the message/otherwise transform the message.  Two suggestions are:
    1. Combine [headerfrom](https://docs.confluent.io/platform/current/connect/transforms/headerfrom.html#headerfrom) with [insertfield](https://docs.confluent.io/platform/current/connect/transforms/insertfield.html#insertfield) to extract the desired part of your key into the header and then insert it into your value
[...]

PR #878, opened 2022-09-01
Add JsonFormatStreamReader

iosifnicolae2:


iosifnicolae2:
Fixes https://github.com/lensesio/stream-reactor/issues/874
Partially fixes https://github.com/lensesio/stream-reactor/issues/869

fqlx:
@andmarios Could someone your team review this?  I need this change as well to backup everything from kafka (timestamp, key, value, headers, and message)

davidsloan (changes requested):
Sorry for taking a while to look at this one.

A few comments:

* Please remove "CONTRIBUTING" file - we will be adding something similar soon.  Also not appropriate for us to have any mention of other companies.
* Addition of Dockerfile looks good but please can you document how this is used or how this can help development in the README.
* I don't think the MANIFEST.MF is necessary as this should be generated by the assembly process? Is there a reason you needed to add this?
* Any code addition should be covered with unit tests so as to ensure the code is bug-free and cover against future regressions.  We plan on covering this in some contribution guidelines in future, apologies for not making this clear.

davidsloan:
I forgot to mention, thank you for the contribution @iosifnicolae2 👍

iosifnicolae2:
@davidsloan thank you for your review, unfortunetly we're not using anymore this functionality so I cannot allocate more time on fixing all the requested changes.

@fqlx  if you have time feel free to fix the requested changes and open a new PR

Thank you!

PR #891, opened 2022-11-20
Fix/assembly pack

andmarios:


andrewstevenson (approved):

project/plugins.sbt
```diff
@@ -11,3 +11,5 @@ addSbtPlugin("net.vonbuchholtz"    % "sbt-dependency-check" % "4.0.0")
 libraryDependencies += "org.slf4j" % "slf4j-nop"            % "1.7.36"
 
 addSbtPlugin("com.eed3si9n" % "sbt-projectmatrix" % "0.9.0")
+
+addSbtPlugin("org.xerial.sbt" % "sbt-pack" % "0.14")
```
davidsloan, line 15: There is a version 0.16 available in maven, would it be best to use the latest?

PR #906, opened 2023-01-11
adding test and Null conversion

eliax1996:
During the deployment of a mongo cdc using the `MongoSourceConnector` we found that certain generated records where containing null elements in the generated array. 
This caused the S3 Sink connector to fail and we managed to solve the problem by patching the connector with the provided conversion to our fork. 
Porting here the fix that we used


davidsloan (approved):
This is great, and thank you for adding the test case too.

Issue #839, opened 2022-04-04
MQTT Source connector throwing NoClassDefinition Found Error when processing message


anuragp89:
Hi,

I am trying to deploy Kafka Connect with MQTT Source connector on Kubernetes.  I am using Strimzi operator to orchestrate the deployment. I have done the deployment and MQTT Source connector is able to connect to MQTT broker, but every time it reads mqtt message, it throws below error:

2022-04-04 09:48:17,924 WARN [kafka-mqtt-connector|task-0] Connection lost. Re-connecting is set to true (com.datamountaineer.streamreactor.connect.mqtt.source.MqttManager) [MQTT Call: paho9676402437775335]
MqttException (0) - java.lang.NoClassDefFoundError: Could not initialize class com.datamountaineer.streamreactor.connect.converters.MsgKey$
        at org.eclipse.paho.client.mqttv3.internal.CommsCallback.run(CommsCallback.java:220)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.NoClassDefFoundError: Could not initialize class com.datamountaineer.streamreactor.connect.converters.MsgKey$
        at com.datamountaineer.streamreactor.connect.converters.source.JsonSimpleConverter.convert(JsonSimpleConverter.scala:56)
        at com.datamountaineer.streamreactor.connect.mqtt.source.MqttManager.messageArrived(MqttManager.scala:103)
        at org.eclipse.paho.client.mqttv3.internal.CommsCallback.deliverMessage(CommsCallback.java:499)
        at org.eclipse.paho.client.mqttv3.internal.CommsCallback.handleMessage(CommsCallback.java:402)
        at org.eclipse.paho.client.mqttv3.internal.CommsCallback.run(CommsCallback.java:206)
        ... 6 more
[...]

PR #913, opened 2023-01-31
Fix for NoClassDefFoundError caused by lack of avro4s libraries in the jar produced by sbt assembly

AnatolyPopov:
A partial fix for https://github.com/lensesio/stream-reactor/issues/839

One of the reasons for NoClassDefFoundError in MQTT connector is caused by the filtering of dependencies to be included in jar produced by `sbt assembly`. 

Since the filtering is based on `String.contains` and the list of excluded patterns include `avro` all the `avro4s` jars are also excluded despite the fact that they are needed at runtime.

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -204,7 +204,7 @@ object Settings extends Dependencies {
           assembly / assemblyExcludedJars := {
             val cp: Classpath = (assembly / fullClasspath).value
             cp filter { f =>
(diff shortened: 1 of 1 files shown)


davidsloan:
Hi thanks for the contribution!  However I believe the code that already made it onto `master` branch should solve this problem as we are now managing dependencies differently.  Let me know if it doesn't.

PR #933, opened 2023-04-12
Upgrade to Influxdb2 client

davidsloan:
Upgrading to use the InfluxDB2 client.

Note: This doesn't yet support Influxdb2 connections but uses the InfluxDB1 compatibility.
This lays the groundwork for supporting both InfluxDB releases within the same module.

### DIFF
--- a/kafka-connect-influxdb/src/main/scala/com/datamountaineer/streamreactor/connect/influx/converters/InfluxPoint.scala
+++ b/kafka-connect-influxdb/src/main/scala/com/datamountaineer/streamreactor/connect/influx/converters/InfluxPoint.scala
@@ -23,7 +23,8 @@ import com.datamountaineer.streamreactor.connect.influx.writers.KcqlDetails
 import com.datamountaineer.streamreactor.connect.influx.writers.KcqlDetails.ConstantTag
 import com.datamountaineer.streamreactor.connect.influx.writers.KcqlDetails.DynamicTag
 import com.datamountaineer.streamreactor.connect.influx.writers.KcqlDetails.Path
-import org.influxdb.dto.Point
+import com.influxdb.client.domain.WritePrecision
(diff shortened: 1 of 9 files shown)


kafka-connect-influxdb/src/main/scala/com/datamountaineer/streamreactor/connect/influx/converters/InfluxPoint.scala
```diff
@@ -135,14 +144,14 @@ object InfluxPoint {
...
-    case value: Int                  => Try(builder.addField(field, value))
+    case value: Int                  => Try(builder.addField(field, value.toLong))
     case value: BigInt               => Try(builder.addField(field, value))
-    case value: Byte                 => Try(builder.addField(field, value.toShort))
-    case value: Short                => Try(builder.addField(field, value))
+    case value: Byte                 => Try(builder.addField(field, value.toLong))
+    case value: Short                => Try(builder.addField(field, value.toLong))
     case value: Double               => Try(builder.addField(field, value))
-    case value: Float                => Try(builder.addField(field, value))
+    case value: Float                => Try(builder.addField(field, value.toDouble))
```
stheppi, line 154: it's this going to break compatibility when upgrading an existing connector?

stheppi (approved):

kafka-connect-influxdb/src/main/scala/com/datamountaineer/streamreactor/connect/influx/converters/InfluxPoint.scala
davidsloan: No, as there have been no changes to the influxdb types - only the mapping which is done invisibly by the Java library.  As this is an increase in precision there will be no impact except this explicit conversion in the connector code.

PR #934, opened 2023-04-25
Connector Logging Configuration

davidsloan:
Allow configuring logging via connector configuration - (currently S3 connector only)

This introduces a new property for the AWS S3 connector:

`connect.s3.log.level.override`

The values this can take are: `DEFAULT`, `OFF`, `ERROR`, `WARN`, `INFO`, `DEBUG`, `TRACE`, `ALL`.  If not specified `DEFAULT` will be used and the logging behaviour will be unchanged from previous releases.

This impacts the log levels of the connector code (anything under the packages `io.lenses` and `com.datamountaineer`).

This allows configuration of the logging levels for systems eg MSK where the Kafka Connect logging API is not available.

### DIFF
--- a/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3ConfigDef.scala
+++ b/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3ConfigDef.scala
@@ -16,6 +16,7 @@
 package io.lenses.streamreactor.connect.aws.s3.config
 
 import cats.implicits.catsSyntaxEitherId
+import cats.implicits.toBifunctorOps
 import com.datamountaineer.streamreactor.common.config.base.traits._
 import com.typesafe.scalalogging.LazyLogging
 import io.lenses.streamreactor.connect.aws.s3.config.processors.ConfigDefProcessor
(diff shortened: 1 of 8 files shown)


kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3ConfigDef.scala
```diff
@@ -297,7 +305,10 @@ case class S3ConfigDefBuilder(sinkName: Option[String], props: util.Map[String,
     with ConnectionSettings
     with S3FlushSettings
     with CompressionCodecSettings
-    with PaddingStrategySettings {
+    with PaddingStrategySettings
+    with LogOverrideSettings {
+
+  configureLoggingOverrides().leftMap(throw _)
```
stheppi, line 311: 🥇

kafka-connect-common/src/main/scala/com/datamountaineer/streamreactor/common/config/base/traits/LogOverrideSettings.scala
```diff
@@ -0,0 +1,45 @@
...
+import com.datamountaineer.streamreactor.common.config.base.const.TraitConfigConst._
+import org.apache.kafka.connect.errors.ConnectException
+import org.slf4j.LoggerFactory
+
+import scala.util.Try
+
+trait LogOverrideSettings extends BaseSettings {
+
+  private val CONNECTOR_LOG_LEVEL_OVERRIDE = s"$connectorPrefix.$LOG_LEVEL_OVERRIDE"
+  private val PACKAGES                     = Set("io.lenses", "com.datamountaineer")
```
stheppi, line 29: Minor. Scala code would say to use Packages and ConenctorLogLevelOverride (I know that Java takes this approach, but even then these are not static)

davidsloan: I would like to address all these style violations en masse.  But probably after some larger pieces of work that are in progress have gone in.  At the moment I will leave it to avoid introducing inconsistencies into the codebase.

PR #937, opened 2023-05-03
Partition Discovery on S3 Source

davidsloan:
This is a large PR that aims to address the issues the S3 Source currently has with discovering files within partitions.

In addition it enables the Source to run independently on multiple workers without duplicating records.  On startup the node is assigned a TASK ID which is used with a hashing algorithm to divide new partitions between the multiple workers.

New Source properties are introduced in order to control the new behaviour:

| Property | Description | 
|--------|--------|
| `connect.s3.partition.search.recurse.levels` | When searching for new partitions on the S3 filesystem, how many levels deep to recurse. |
|  `connect.s3.partition.search.block` | 'true' will block the thread, 'false' will search for new partitions in the background. |
|  `connect.s3.partition.search.interval` | The interval in milliseconds between searching for new partitions.  Defaults to 5 minutes. |
|  `connect.s3.partition.search.pause.after.count` | When searching for new partitions, how many partitions to find before pausing.  Defaults to -1 (completes the search - no pause). |
|  `connect.s3.partition.search.pause.after.time` | When searching for new partitions, how long to run the search for before pausing.  Defaults to -1 (completes the search - no pause). |

In addition there has been a large amount of refactoring.  This includes:
* Organising the packages between Source and Sink.
* Splitting the `ConfigDef` into `S3SinkConfigDef` and `S3SourceConfigDef` as having them merged made no sense.


kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/S3SinkConfigDef.scala
```diff
@@ -288,8 +133,21 @@ class S3ConfigDef() extends ConfigDef with LazyLogging {
...
-    extends BaseConfig(S3ConfigSettings.CONNECTOR_PREFIX, S3ConfigDef.config, props)
+object S3SinkConfigDefBuilder {
+  def apply(
+    props:        util.Map[String, String],
+    contextProps: util.Map[String, String],
+  ): S3SinkConfigDefBuilder =
+    S3SinkConfigDefBuilder(
+      Option(contextProps)
+        .filterNot(_.isEmpty)
+        .getOrElse(props),
```
stheppi, line 144: can `props` be null? 

Minor: it would help to name them props + fallback/default. This might instruct what this does.

stheppi (approved):

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/state/OpenS3SourceTaskState.scala
```diff
@@ -0,0 +1,100 @@
...
+  ): Either[Throwable, S3SourceTaskState] = new ConnectException("Cannot start").asLeft
+
+}
+
+object OpenS3SourceTaskState {
+  def apply(
+    props: util.Map[String, String],
+  )(
+    implicit
+    contextPropsFn:  () => util.Map[String, String],
```
stheppi, line 64: Minor: Is there a reason we are not passing the functions "manually" rather than implicitly?  Just for visibility on immediate argumnets.

stheppi: Furthermore, on implicits: it is unclear why a Clock instance is needed. In GitHub is hard to see if the code needs that implicitly. Otherwise, seeing Clock.defaultTimezone misleads thinking the Clock instance is unnecessary.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/reader/ReaderManagerService.scala
```diff
@@ -0,0 +1,89 @@
...
+    if (rediscoverDue && settings.blockOnSearch) {
+      rediscover()
+    } else if (rediscoverDue && !settings.blockOnSearch) {
+      rediscover().start *> IO.unit
+    } else {
+      IO.unit
+    }
+  }
+
+  private def rediscover(): IO[Unit] =
```
stheppi, line 65: using a Ref, and considering there is one writer only I would suggest something like:

```
IO[ReaderManagerState] =
    for {
      _        <- IO(logger.debug("calculating updated state"))
      ioState  <- ref.get
      newParts <- partitionSearcher.findNewPartitions(ioState.partitionResponses)
      newReaderManagers = newParts
        .flatMap(part => part.results.partitions.map(part.root -> _))
        .map {
          case (location, res) => readerManagerCreateFn(location, res)
        }
        newState = ioState.copy(
      partitionResponses = newParts,
      readerManagers     = ioState.readerManagers ++ newReaderManagers,
    )
     _ <- refState.set(newState)
    } yield ()
```

I am not sure if it compiles. Happy to go on a call. 

This solution avoids creating an IO to get the value of the AtomicReference and invoke unsafeRunSync

kafka-connect-aws-s3/src/test/scala/io/lenses/streamreactor/connect/aws/s3/model/HierarchicalPartitionExtractorTest.scala
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.lenses.streamreactor.connect.aws.s3.model
+
+import com.typesafe.scalalogging.LazyLogging
+import io.lenses.streamreactor.connect.aws.s3.source.config.HierarchicalPartitionExtractor
+import org.scalatest.flatspec.AnyFlatSpecLike
+
+class HierarchicalPartitionExtractorTest extends AnyFlatSpecLike with LazyLogging {
```
stheppi, line 22: Is this missing tests?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3DirectoryLister.scala
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+  private def extractPrefixesFromResponse(
+    listObjectsV2Response: ListObjectsV2Iterable,
+    exclude:               Set[String],
+    completionConfig:      DirectoryFindCompletionConfig,
+  ): IO[DirectoryFindResults] =
+    IO {
+      val rtn: mutable.Set[String] = mutable.Set()
+      val resp = listObjectsV2Response
```
stheppi, line 109: If lostObjectsV2Response is converted to a Scala iterator you can just use 
```
foldLeft(Set.empty){ case (set, item)=> 

...
set ++ commonPRefixesFiltered
}
```

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3DirectoryLister.scala
```diff
@@ -0,0 +1,140 @@
...
+        val listResp: ListObjectsV2Response = resp.next()
+        val commonPrefixesFiltered = listResp
+          .commonPrefixes()
+          .asScala
+          .map(_.prefix())
+          .filter(connectorTask.ownsDir)
+          .filterNot(exclude.contains)
+        rtn.addAll(commonPrefixesFiltered)
+
+        partialResultInfo =
```
stheppi, line 125: If you use foldLeft as I suggested earlier, you can also accumulate the partialResultInfo without resorting to mutability.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/reader/ReaderManagerService.scala
FPerezP: Yeah, I also agree with the usage of Ref here, I already mentioned above in a different file

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3DirectoryLister.scala
```diff
@@ -0,0 +1,140 @@
...
+  private def flattenPrefixes(
+    bucketAndPrefix:  RemoteS3RootLocation,
+    prefixes:         Set[String],
+    completionConfig: DirectoryFindCompletionConfig,
+    exclude:          Set[String],
+  )(
+    implicit
+    clock: Clock,
+  ): IO[Set[String]] =
+    if (completionConfig.levelsToRecurse == 0) IO.pure(prefixes)
```
FPerezP, line 66: I would use `.delay` instead of `.pure` in general. The main issue with pure is that the passed param is eagerly evaluated, so any error produced there would raise the exception out of the IO/F context. 
We can use the regular apply for IO wich already delagates the construction to the delay method.
This is especially interesting/important when dealing with a Java api, since exceptions are part of their nature :)

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3Config.scala
```diff
@@ -102,6 +113,13 @@ object FormatOptions extends Enum[FormatOptions] {
 
 case object FormatSelection {
 
+  def fromKcql(
+    kcql: Kcql,
+  ): FormatSelection =
+    Option(kcql.getStoredAs) match {
```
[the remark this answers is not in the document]
davidsloan: Done!

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/S3SinkConfigDef.scala
davidsloan: Refactored to make this behaviour clearer and further protect against nulls.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/reader/ReaderManagerService.scala
```diff
@@ -0,0 +1,89 @@
...
+
+}
+
+class ReaderManagerService(
+  settings:              PartitionSearcherOptions,
+  partitionSearcher:     PartitionSearcher,
+  readerManagerCreateFn: (RemoteS3RootLocation, String) => ReaderManager,
+) extends LazyLogging {
+
+  private val readerManagerState: AtomicReference[ReaderManagerState] =
```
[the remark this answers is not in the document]
davidsloan: Changed.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/reader/ReaderManagerService.scala
davidsloan: Fixed!

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3DirectoryLister.scala
davidsloan: I tried this however this doesn't quite here because we can't loop discriminately through the whole of the list as the iterator hides paging functionality (calling more pages, which may not be desired depending on the configuration).  The way it is implemented we are able to pause the process before we iterate through every item.

I would like a better way of doing this, however as I appreciate it is not too elegant.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/TaskDistributor.scala
```diff
@@ -0,0 +1,30 @@
...
+import io.lenses.streamreactor.connect.aws.s3.config.S3ConfigSettings.TASK_INDEX
+
+import java.util
+import scala.jdk.CollectionConverters.SeqHasAsJava
+
+object TaskDistributor {
+  def distributeTasks(props: util.Map[String, String], maxTasks: Int): util.List[util.Map[String, String]] =
+    List.tabulate(maxTasks) { i =>
+      val map: util.Map[String, String] = new util.HashMap[String, String](props)
+      map.put(TASK_INDEX, s"$i:$maxTasks")
```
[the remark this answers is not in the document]
davidsloan: This has to be a Java HashMap as it is used to pass the TASK_INDEX and other properties from one connect worker to the other and is serialised by the underlying Connect framework.  It would likely not handle the Scala map type.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/LocalStagingArea.scala
```diff
@@ -25,18 +25,23 @@ import java.util.UUID
...
+  def apply(
+    confDef: S3SinkConfigDefBuilder,
+  )(
+    implicit
+    connectorTaskId: ConnectorTaskId,
+  ): Either[Exception, LocalStagingArea] =
     getStringValue(confDef.getParsedValues, LOCAL_TMP_DIRECTORY)
       .orElse {
-        confDef.sinkName.map(createLocalDirBasedOnSinkName)
+        Option.when(!connectorTaskId.hasDefaultConnectorName)(createLocalDirBasedOnSinkName)
```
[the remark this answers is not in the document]
davidsloan: Refactored this code to achieve your suggestion

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/reader/ReaderCreator.scala
```diff
@@ -15,37 +15,39 @@
...
 import io.lenses.streamreactor.connect.aws.s3.storage.StorageInterface
 import org.apache.kafka.common.errors.OffsetOutOfRangeException
 
 import scala.util.Try
 
 class ReaderCreator(
-  sourceName:  String,
   format:      FormatSelection,
   targetTopic: String,
+  partitionFn: String => Option[Int],
```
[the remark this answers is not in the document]
davidsloan: Let me know if you find anything.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTask.scala
```diff
@@ -41,44 +44,37 @@ import scala.util.Try
...
 class S3SinkTask extends SinkTask with ErrorHandler {
 
+  implicit val contextPropsFn: () => util.Map[String, String] = SinkContextReader.getProps(() => context)
+
   private val manifest = JarManifest(getClass.getProtectionDomain.getCodeSource.getLocation)
 
   private var writerManager: S3WriterManager = _
 
-  private var sinkName: String = _
+  private implicit var connectorTaskId: ConnectorTaskId = DefaultConnectorTaskId
```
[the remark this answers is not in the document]
davidsloan: Unfortunately we are in a Java API that requires state here.  The `start` method has to set up the state based on the properties.  I can't think of a way to avoid these.

To avoid the mutability in the past I have introduced finite state models into the code.  This doesn't completely remove the mutability as you still have to have a variable to keep your state in.  If you have any ideas I'd be happy to hear them.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3Config.scala
```diff
@@ -35,7 +37,12 @@ import scala.collection.immutable
 import scala.util.Try
 
 sealed trait AwsClient extends EnumEntry {
-  def createStorageInterface(connectorName: String, authResources: AuthResources): Either[String, StorageInterface]
+  def createStorageInterface(
+    authResources: AuthResources,
+  )(
+    implicit
+    connectorTaskId: ConnectorTaskId,
```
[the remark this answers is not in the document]
davidsloan: Thanks for the comment Fran!  As we discussed, but I'm recording here:

The field is immutable, it is effectively a global value (it is used many places).
It is effectively sourced from a piece of configuration that will not change.
The field is required by much of the code as an ancillary (in most cases with the classes that use it, it is not the primary concern of the class in question)
To increase the brevity of the code, as an implicit.

For these reasons we discussed I think I would like to keep it as an implicit.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/distribution/PartitionSearcher.scala
```diff
@@ -0,0 +1,94 @@
...
+         lastFound.map {
+           prevResponse =>
+             findNewPartitionsInRoot(
+               prevResponse.root,
+               settings,
+               prevResponse.allPartitions,
+               resumeFrom(prevResponse),
+             )
+         }
+       }).sequence
```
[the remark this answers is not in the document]
davidsloan: A really good suggestion, I have updated the instances using `.sequence` to use `.traverse` instead`

kafka-connect-aws-s3/src/test/scala/io/lenses/streamreactor/connect/aws/s3/model/HierarchicalPartitionExtractorTest.scala
davidsloan: Added!

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/SinkContextReader.scala
```diff
@@ -0,0 +1,37 @@
...
+import scala.jdk.CollectionConverters.MapHasAsScala
+import scala.util.Try
+
+object SinkContextReader {
+
+  def mergeProps(
+    contextPropsFn: () => util.Map[String, String],
+  )(fallbackProps:  util.Map[String, String],
+  ): Either[String, util.Map[String, String]] = {
+    val context  = Try(contextPropsFn()).toOption.getOrElse(Collections.emptyMap())
```
stheppi, line 32: Please help me understand the context a bit. Maybe adding a comment will help further. 
When would a config builder (contextPropsFn) fail? 

Why would the fallbackProps be null?
We always return success so is there any value having Either[String, util.Map...]?

davidsloan: I think this was in response to your comment:

https://github.com/lensesio/stream-reactor/pull/937#discussion_r1184726811

davidsloan: > When would a config builder (contextPropsFn) fail?

1. Internal error in Kafka connect where the task id is not recognised.  Very unlikely.
2. When we write integration tests but do not want to mock the context every time. (real answer)

> Why would the fallbackProps be null?
I don't think they would.

> We always return success so is there any value having Either[String, util.Map...]?
Good point.  Consider it gone.

FPerezP (approved):

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/config/PartitionSearcherOptions.scala
```diff
@@ -0,0 +1,39 @@
...
+case class PartitionSearcherOptions(
+  recurseLevels:               Int,
+  blockOnSearch:               Boolean,
+  searchInterval:              Duration, // searches again or resumes search in partial mode
+  pauseSearchOnPartitionCount: Option[Int], // this is per root
+  pauseSearchAfterTime:        Option[Duration], // this is per root
+  maybeClock:                  Option[Clock] = None,
+) {
+
+  private val clock = maybeClock.getOrElse(Clock.systemDefaultZone())
```
stheppi, line 31: why not let the caller provide the Clock instance. I think it is better to avoid these defaults and expose them

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/files/S3SourceLister.scala
```diff
@@ -16,21 +16,23 @@
...
   def listBatch(
     bucketAndPrefix: RemoteS3RootLocation,
     lastFile:        Option[RemoteS3PathLocation],
     numResults:      Int,
+  )(
+    implicit
+    storageInterface: StorageInterface,
   ): Either[Throwable, List[RemoteS3PathLocation]] =
-    sourceStorageInterface.list(bucketAndPrefix, lastFile, numResults).map(_.map(bucketAndPrefix.withPath))
+    storageInterface.list(bucketAndPrefix, lastFile, numResults).map(_.map(bucketAndPrefix.withPath))
```
stheppi, line 37: Isn't this going to be a problem if we do not use the page-based listing? if the bucket has 100K files the call will potentially fail

davidsloan: It will pull out a max of 1000 filenames by default per call (aws limits).  It only needs to get the next file, the rest are stored in the queue for future use.  You will notice that the last file seen is always passed in, we resume after this point.

davidsloan: while investigating your comment I noticed that this class wasn't doing anything, so it is gone.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTask.scala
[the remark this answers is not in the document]
stheppi: From a concurrency perspective we have one writer : the framework itself calling start. Start won't be called more than once per task instance. Since it is a java API and the task needs to track its own state, we cannot avoid the var.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/reader/ReaderManagerService.scala
```diff
@@ -0,0 +1,84 @@
...
+  private def launchDiscover(): IO[Unit] = {
+    for {
+      ioState       <- readerManagerState.get
+      lastSearchTime = ioState.lastSearchTime
+      rediscoverDue  = settings.rediscoverDue(lastSearchTime)
+    } yield {
+      if (rediscoverDue && settings.blockOnSearch) {
+        rediscover()
+      } else if (rediscoverDue && !settings.blockOnSearch) {
+        rediscover().start *> IO.unit
```
stheppi, line 60: It is best to use .background here.

stheppi (approved):

stheppi (approved):

PR #943, opened 2023-05-19
S3 Sink Connector - Remove JClouds, Remove Legacy Offset Seek

davidsloan:
This removes some deprecated features from the S3 Sink Connector:
* jclouds client support - this reduces the footprint of the distributable and there were issues reported with the jclouds functionality which is why we have replaced it with the official AWS client.  From the next release, the `connect.s3.aws.region` will be mandatory for the S3 Sink if not configuring using another authentication method.
* legacy offset seek support - anyone using this should have migrated to the new offset seeker.  If you haven't please use 4.2.x to migrate before upgrading.


stheppi (approved):
lgtm

PR #945, opened 2023-05-22
Parquet INT96 support

stheppi:
Allows Parquet INT96 to be read as a fixed size array. Otherwise, it fails at runtime.

Updates the dependency on parquet to 1.13.1


davidsloan (approved):
Looks good.

PR #947, opened 2023-05-31
Feat/improvements rebase

stheppi:
#### Description

Brings code improvements 

* avoid the use of unsafeRunSync in the code apart from the source task where the interaction with the Java API happens
* removes Open/Close state to keep the state in the source task. The code relies now on Option to specify not-initialised or initialised.
* Removes AuthResource.scala since it has two indirections to call new AwsS3ClientCreator.createS3Client. The trait is not used, creating an instance of AuthResources to call the `aws` method can be avoided with one builder function
* Removes ThrowableEither.scala and changes the code where the Either[String, T] was used to Either[Throwable, T], thus improving the code flow and a few for-comprehension
* Readercreator is an instance builder for ResultReader, thus the code was moved to ResultReader companion-object 
* connectorTaskId used in logging is moved as normal param, not implicit
* small improvements by avoiding allocations by copying immutable objects during some for-comprehension
* S3Location removes pathOrError. It reduces the complexity of navigating a few functions to get the result versus having a leftMap where pathOrError is called.
* S3SourceFileQueue has dropped the markComplete and init. markComplete is redundant if the class is to behave as a queue. Once a file is picked is removed from the internals, and it does not require an external call to read. The init function is not ideal for it to be called at any point. It has been replaced with a specific builder, which sets the initial state to the one initial file.
* ReaderManager has been rewritten to avoid so many states. A file is read, and once it completes, it moves to the next file, if there is one. It reduces the code complexity of having a few state types. Furthermore, the code was changed to use the effect type IO.


FPerezP (approved):
Approved. We already discussed some areas to improve on the call we had earlier today

PR #949, opened 2023-06-05
Decouples the partition discovery from every Connect poll requests.

stheppi:
A background task is created to check on an interval.

The code removes the option to have a partial search or a search which terminates after a time window. There's always a full search executed.


kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/distribution/PartitionSearcher.scala
```diff
@@ -21,36 +21,30 @@ import com.typesafe.scalalogging.LazyLogging
...
+  connectorTaskId: ConnectorTaskId,
+  listS3ObjF:      ListObjectsV2Request => Iterator[ListObjectsV2Response],
 ) extends LazyLogging {
 
   def findNewPartitions(
     lastFound: Seq[PartitionSearcherResponse],
   ): IO[Seq[PartitionSearcherResponse]] =
     if (lastFound.isEmpty && roots.nonEmpty) {
-      roots.traverse(findNewPartitionsInRoot(_, settings, Set.empty, Option.empty, settings.clock))
+      roots.traverse(findNewPartitionsInRoot(_, settings, Set.empty))
```
FPerezP, line 40: Not part of this PR. Using traverse here would make findNewPartitions fully fail for any single failing element inside root.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/reader/PartitionDiscovery.scala
```diff
@@ -0,0 +1,63 @@
...
+      oldState <- readerManagerState.get
+      newParts <- partitionSearcher.findNewPartitions(oldState.partitionResponses)
+      tuples    = newParts.flatMap(part => part.results.partitions.map(part.root -> _))
+      newReaderManagers <- tuples
+        .map { case (location, path) => readerManagerCreateFn(location, path) }.traverse(identity)
+      newState = oldState.copy(
+        partitionResponses = newParts,
+        readerManagers     = oldState.readerManagers ++ newReaderManagers,
+      )
+      _ <- readerManagerState.set(newState)
```
FPerezP, line 60: Could we have multiple parallel calls to the `run` method so that we ended up overriding the state, or calculating the new state based on an out of date one?

FPerezP (approved):
Minor comments. Approved :rocket:

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/distribution/PartitionSearcher.scala
stheppi: That's a desired behaviour.

PR #952, opened 2023-06-07
Format Refactor / StartEndTagTextReader / RegexTextReader Wiring

davidsloan:
Refactor out most usages of Format in favour of FormatSelection.  Move config around a little.  Wire in the StartEndTagTextReader and RegexTextReader via an adaptor.

### DIFF
--- a/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/FormatSelection.scala
+++ b/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/FormatSelection.scala
@@ -131,7 +131,8 @@ case object ParquetFormatSelection extends FormatSelection {
     fileSize:             Long,
     bucketAndPath:        S3Location,
     recreateInputStreamF: () => Either[Throwable, InputStream],
-  ) = ParquetFormatStreamReader.apply(inputStream, fileSize, bucketAndPath, recreateInputStreamF)
+  ): ParquetFormatStreamReader =
+    ParquetFormatStreamReader.apply(inputStream, fileSize, bucketAndPath, recreateInputStreamF)
 
   override def extension: String = "parquet"
 }
@@ -142,13 +143,7 @@ case class TextFormatSelection(readTextMode: Option[ReadTextMode]) extends Forma
     bucketAndPath:        S3Location,
     recreateInputStreamF: () => Either[Throwable, InputStream],
   ): S3FormatStreamReader[StringSourceData] =
-    readTextMode.map(_.createStreamReader(inputStream, bucketAndPath))
-      .getOrElse(
-        new TextFormatStreamReader(
-          inputStream,
-          bucketAndPath,
-        ),
-      )
+    TextFormatStreamReader(readTextMode, inputStream, bucketAndPath)
 
   override def extension: String = "text"
 }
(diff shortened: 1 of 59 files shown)


kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/FormatSelection.scala
```diff
@@ -0,0 +1,179 @@
...
+  override def extension: String = "parquet"
+}
+case class TextFormatSelection(readTextMode: Option[ReadTextMode]) extends FormatSelection {
+  override def toStreamReader(
+    inputStream:          InputStream,
+    fileSize:             Long,
+    bucketAndPath:        S3Location,
+    recreateInputStreamF: () => Either[Throwable, InputStream],
+  ): S3FormatStreamReader[StringSourceData] =
+    readTextMode.map(rtm => rtm.createStreamReader(inputStream, bucketAndPath)).getOrElse(new TextFormatStreamReader(
```
stheppi, line 145: Would it not improve the code to have a TextFromatStreamReader to be built like the other two and avoid this code being responsible for building the readers?

kafka-connect-common/src/main/scala/io/lenses/streamreactor/connect/io/text/OptionIteratorAdaptor.scala
```diff
@@ -0,0 +1,35 @@
...
+ */
+package io.lenses.streamreactor.connect.io.text
+
+class OptionIteratorAdaptor(optional: () => Option[String]) extends Iterator[String] {
+
+  private var lineNumber: Long           = 0L
+  private var nextVal:    Option[String] = Option.empty
+
+  override def hasNext: Boolean = {
+    nextVal = optional()
```
stheppi, line 24: I see this is required to initialise nextVal, and avoid initalisation in ctor. But I think it creates a weird behaviour. An iterator hasNext can be called as many times, it's only next:T which mutates. 

A few points on this:
1. Violation of Single Responsibility Principle: The hasNext() method should focus solely on checking whether there is a next element. Reading the next value in this method would introduce additional responsibilities, violating the principle of having a single responsibility for each method.

2. Side Effects and Advancement of the Iterator: Reading the next value within hasNext() would advance the iterator, altering its state before the client calls next(). This can lead to unexpected behavior and make the code harder to reason about.

3.Performance Considerations: Reading the next value in hasNext() can have performance implications, especially if retrieving the next value involves expensive operations. It is generally more efficient to separate the responsibility of advancing the iterator from checking for the next element.


there are a few options.

1. compute the nextVal in the ctor

```
class OptionIteratorAdaptor(optional: () => Option[String]) extends Iterator[String] {

  private var lineNumber: Long           = 0L
  private var nextVal:    Option[String] = optional()

  override def hasNext: Boolean =  nextVal.isDefined
  
  def next(): String = nextVal match {
    case Some(element) =>
      val currentElement = element
      nextVale = optional()
      currentElement
    case None => //raise an error
  }
```

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/FormatSelection.scala
```diff
@@ -0,0 +1,179 @@
...
+      inputStream,
+      bucketAndPath,
+    ))
+
+  override def extension: String = "text"
```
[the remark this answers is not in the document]
davidsloan: Maintaining existing behaviour, although it makes total sense, we can't change this now 😨

kafka-connect-common/src/main/scala/io/lenses/streamreactor/connect/io/text/OptionIteratorAdaptor.scala
davidsloan: Thank you for the advice and the explanation 🙇

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/FormatSelection.scala
davidsloan: I have moved this.  Thanks!

stheppi (approved):

PR #954, opened 2023-06-09
Fix/s3 source

stheppi:


kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3DirectoryLister.scala
```diff
@@ -77,8 +78,9 @@ object AwsS3DirectoryLister {
       .maxKeys(1000)
       .bucket(bucketAndPrefix.bucket)
       .delimiter("/")
-    bucketAndPrefix.prefix.map(addTrailingSlash).foreach(builder.prefix)
-    builder.build()
+    bucketAndPrefix.prefix.foreach(builder.prefix)
+    val request = builder.build()
```
davidsloan, line 82: minor, but no need for the 'request' variable as you can just have `builder.build()` as the last line.

kafka-connect-aws-s3/src/test/scala/io/lenses/streamreactor/connect/aws/s3/source/config/ReadTextModeTestFormatSelection.scala
```diff
@@ -32,7 +34,20 @@ class ReadTextModeTestFormatSelection extends AnyFlatSpec with Matchers {
           S3PropsKeyEnum.ReadEndTag.entryName   -> "</p>",
         ),
       ),
-    ) should be(Some(StartEndTagReadTextMode("<p>", "</p>")))
+    ) should be(Some(StartEndTagReadTextMode("<p>", "</p>", 1024)))
```
davidsloan, line 37: could the 1024 be a variable with a name or comment that explains the magic number?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/FormatSelection.scala
```diff
@@ -46,19 +47,18 @@ case object FormatSelection {
...
-      .map(FormatSelection.fromString(_, readTextModeFn))
-      .getOrElse(JsonFormatSelection)
-  }
+  ): Either[Throwable, FormatSelection] =
+    Option(kcql.getStoredAs) match {
+      case Some(storedAs) =>
+        fromString(storedAs, () => ReadTextMode(schema.readProps(kcql.getProperties.asScala.toMap)))
+      case None =>
+        Right(JsonFormatSelection)
+    }
```
davidsloan, line 56: Is it nicer to avoid the case match by doing something like this?

```suggestion
    Option(kcql.getStoredAs)
      .map(fromString(_, () => ReadTextMode(schema.readProps(kcql.getProperties.asScala.toMap))))
      .getOrElse(JsonFormatSelection)
```

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/FormatSelection.scala
```diff
@@ -68,15 +68,18 @@ case object FormatSelection {
...
+          case Format.Json    => JsonFormatSelection
+          case Format.Avro    => AvroFormatSelection
+          case Format.Parquet => ParquetFormatSelection
+          case Format.Text    => TextFormatSelection(readTextMode())
+          case Format.Csv     => CsvFormatSelection(formatOptions)
+          case Format.Bytes   => BytesFormatSelection(formatOptions)
+        }
+        result.asRight
+      case None => Left(new IllegalArgumentException(s"Unsupported format - $formatAsString"))
     }
```
davidsloan, line 83:
```suggestion
    
    Format
      .withNameInsensitiveOption(split(0))
      .map {
        case Format.Json    => JsonFormatSelection
        case Format.Avro    => AvroFormatSelection
        case Format.Parquet => ParquetFormatSelection
        case Format.Text    => TextFormatSelection(readTextMode())
        case Format.Csv     => CsvFormatSelection(formatOptions)
        case Format.Bytes   => BytesFormatSelection(formatOptions)
      }
      .toRight(new IllegalArgumentException(s"Unsupported format - $formatAsString"))
```

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/reader/ResultReader.scala
```diff
@@ -98,7 +98,7 @@ object ResultReader extends LazyLogging {
       inputStream <- storageInterface.getBlob(pathWithLine.bucket, path).leftMap(_.toException)
       fileSize    <- storageInterface.getBlobSize(pathWithLine.bucket, path).leftMap(_.toException)
       _ <- Try(logger.info(
-        s"[${connectorTaskId.show}] Reading next file: ${pathWithLine.show} from line ${pathWithLine.line}",
+        s"[${connectorTaskId.show}] Reading next file: ${pathWithLine.toPath} from line ${pathWithLine.line}",
```
davidsloan, line 101: I notice the following:

S3Location.toPath is only ever used in loggers
S3Location.toString has been replaced with show
S3Location.fromStart is no longer used

Can these 3 methods be removed and only use `pathWithLine.show` going forward?

davidsloan (commented):
A few minor points.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/FormatSelection.scala
stheppi: The method returns an Either not an option. `fromstring` returns an Either, so it's changing the "container", so that map does not work. 

This would be the alternative to avoid the pattern match, but I don't think it reads faster.

```    Option(kcql.getStoredAs).fold(JsonFormatSelection.asInstanceOf[FormatSelection].asRight[Throwable]) { storedAs =>
      fromString(storedAs, () => ReadTextMode(schema.readProps(kcql.getProperties.asScala.toMap)))
    }
    ```

File: kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/reader/ResultReader.scala (hunk 5 above)
Comment (stheppi) [reply kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/reader/ResultReader.scala:101]: Valid point, however if you look at the `show` method

```
    Show.show(loc => s"${loc.bucket}//:${loc.prefix.map(s => s"$s/").getOrElse("")}|${loc.path.getOrElse("")}")
```
It uses `|` and that does not make it quick to use. 

The `show` does not provide the full S3Location, and a toString should capture all info. Whereas we need most of the time the path only.

stheppi: I removed the path, but i think there needs to be a distinction between show and tostring, only because different info is required. Show is meant to be used as a better .tostring, and we invert the usage here, for now.

davidsloan (approved):

Issue #956, opened 2023-06-12
AWS S3 Sink fails to remove index files for GCP Cloud Storage via the S3 compatible APIs


jdfoster:
I am investigating using the AWS S3 Sink to write to a GCP Cloud Storage using the S3-compatible APIs ([docs](https://cloud.google.com/storage/docs/interoperability)). Writing to GCP has been partially successful; I have managed to get the connector to write files, but it cannot delete the index files. Unfortunately, the S3-compatible APIs do not currently support the `deleteObjects` methods ([issue](https://issuetracker.google.com/issues/162653700)).

I have resolved the error by patching the `deleteFiles function` of the `AwsS3StorageInterface` class to loop over the `deleteObject` method. While I understand we are using an unsupported object store, is this an acceptable change to make to the project?

---
# Issue Guidelines

Please review these questions before submitting any issue?

### What version of the Stream Reactor are you reporting this issue for?
`4.2.0`

### Are you running the correct version of Kafka/Confluent for the Stream reactor release?
We are using Kafka `3.3.1` created via the Strimzi operator.

### Do you have a supported version of the data source/sink .i.e Cassandra 3.0.9?
- AWS S3
- GCP Cloud Storage (via the S3-compatible APIs)

### Have you read the [docs](https://docs.lenses.io/connectors/)?
Yes

### What is the expected behaviour?
Messages to be written to the object store bucket.

### What was observed?
Sink process stops as the index files are unable to be deleted.

### What is your Connect cluster configuration (connect-avro-distributed.properties)?
```
```

### What is your connector properties configuration (my-connector.properties)?
```
```

### Please provide full log files (redact and sensitive information)

```
2023-06-09 15:45:10,503 ERROR [***|task-0] [***] Error on cleanup: error deleting file (.indexes/***/***/00000/00000000000000000499 - .indexes/***/***/00000/00000000000000000999 - .indexes/***/***/00000/00000000000000001499 - .indexes/***/***/00000/00000000000000001999) Invalid argument. (Service: S3, Status Code: 400, Request ID: null) (io.lenses.streamreactor.connect.aws.s3.sink.seek.IndexManager) [task-thread-***-0]
software.amazon.awssdk.services.s3.model.S3Exception: Invalid argument. (Service: S3, Status Code: 400, Request ID: null)
[...]

PR #957, opened 2023-06-12
Switch S3 API client call from deleteObjects to deleteObject

jdfoster:
This PR updates the AWS S3 Sink connector to the single object delete method to resolve issue #956. This change allows the sink to work with the S3-compatible APIs for the GCP Cloud Storage.

The S3 client deleteObjects method is unsupported in the GCP Cloud Storage API; this prevents the removal of the index file and blocks the process. By looping over the deleteObject method, we can replicate request using methods supported by GCP Cloud Storage.


davidsloan:
Hi!  Thanks for the contribution and the investigative work you've done on getting this fixed.

I think this is a behaviour that we should add as a connector property.  For example `aws.s3.source.delete.mode = default (all-at-once) / individually`.

This will:
* avoid the performance drop that will negatively impact the 'real' S3 sink
* prevent change in behaviour to users upgrading.
* make sense to have both available as an option.

Also we would need to add tests for these.  We're working on a contribution framework in the future to give a guide on contributions but this isn't ready yet.

Let me know if you need any help implementing this.

jdfoster:
Thanks for taking a look. It makes total sense to gate the change behind a feature flag. Think I understand how to include the configuration flag, but I would appreciate any pointers to implement switching between the two at the AWS Storage Interface layer.

davidsloan:
> Thanks for taking a look. It makes total sense to gate the change behind a feature flag. Think I understand how to include the configuration flag, but I would appreciate any pointers to implement switching between the two at the AWS Storage Interface layer.

No problem at all.

I think I would do the following:

- Add a constructor arg to the AWSStorageInterface (batchDeletes) to receive the configuration
```
class AwsS3StorageInterface(val connectorTaskId: ConnectorTaskId, val s3Client: S3Client, val batchDeletes: Boolean)
```
- In the deleteFiles method dispatch to 2 new methods depending on this value:
```  override def deleteFiles(bucket: String, files: Seq[String]): Either[FileDeleteError, Unit] = Try {
    if (files.isEmpty) {
      return ().asRight
    }
    if(batchDeletes) {
      batchDelete(bucket, files)
    } else {
      loopDelete(bucket, files)
    }

def batchDelete(bucket: String, files: Seq[String])...

def loopDelete(bucket: String, files: Seq[String])...

```
code common to both can stay in the deleteFiles method

Hope this helps!

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3ConfigSettings.scala
```diff
@@ -134,4 +134,9 @@ object S3ConfigSettings {
   val SOURCE_ORDERING_TYPE:         String = s"$CONNECTOR_PREFIX.ordering.type"
   val SOURCE_ORDERING_TYPE_DOC:     String = "AlphaNumeric (the default)"
   val SOURCE_ORDERING_TYPE_DEFAULT: String = "AlphaNumeric"
+
+  val SOURCE_DELETE_MODE         = s"$CONNECTOR_PREFIX.source.delete.mode"
```
jfuller95, line 138: It appears you may need colons after:
```
val SOURCE_DELETE_MODE:
```
for each of the below in this file

kafka-connect-aws-s3/src/test/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/DeleteModeSettingTest.scala
```diff
@@ -0,0 +1,27 @@
...
+class DeleteModeSettingTest extends AnyFlatSpec with Matchers with LazyLogging {
+  it should "respect the delete mode setting" in {
+    val deleteModeMap = Table[String, String, Boolean](
+      ("testName", "value", "expected"),
+      ("all-at-once", "AllAtOnce", true),
+      ("individual", "Individual", false),
+    )
+
+    forAll(deleteModeMap) {
+      (name: String, value: String, expected: Boolean) =>
```
davidsloan, line 19: Move the Table outside of the loop and use it to parameterise the test/set the test name

```suggestion
    
  private val deleteModeMap = Table[String, String, Boolean](
    ("testName", "value", "expected"),
    ("all-at-once", "AllAtOnce", true),
    ("individual", "Individual", false),
  )

  forAll(deleteModeMap) {
    (name: String, value: String, expected: Boolean) =>
    it should s"respect the delete mode setting $name" in {
```

kafka-connect-aws-s3/src/test/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/DeleteModeSettingTest.scala
```diff
@@ -0,0 +1,27 @@
+package io.lenses.streamreactor.connect.aws.s3.sink.config
+
+import com.typesafe.scalalogging.LazyLogging
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.prop.TableDrivenPropertyChecks._
+
+import scala.jdk.CollectionConverters.MapHasAsJava
+
+class DeleteModeSettingTest extends AnyFlatSpec with Matchers with LazyLogging {
```
davidsloan, line 10: To match the name of the class it's testing, this should be `DeleteModeSettingsTest`

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3StorageInterface.scala
```diff
@@ -34,7 +34,7 @@ import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
 
-class AwsS3StorageInterface(val connectorTaskId: ConnectorTaskId, val s3Client: S3Client)
+class AwsS3StorageInterface(val connectorTaskId: ConnectorTaskId, val s3Client: S3Client, val batchDelete: Boolean)
```
davidsloan, line 37: writing any unit tests for this class would be appreciated, but totally optional as I realise there aren't currently any.

davidsloan (commented):
Looks good, a few minor things to work on.

davidsloan:
Also, after you're done then do a `sbt formatAll` then it will take care of formatting for you.

jdfoster:
Just getting back to this now. I have made the changes you asked for.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3StorageInterface.scala
jdfoster: Do you have an example I could copy?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/state/S3SourceBuilder.scala
```diff
@@ -42,7 +42,7 @@ object S3SourceState {
       connectorTaskId  <- IO.fromEither(ConnectorTaskId.fromProps(props))
       config           <- IO.fromEither(S3SourceConfig.fromProps(props))
       s3Client         <- IO.fromEither(AwsS3ClientCreator.make(config.s3Config))
-      storageInterface <- IO.delay(new AwsS3StorageInterface(connectorTaskId, s3Client))
+      storageInterface <- IO.delay(new AwsS3StorageInterface(connectorTaskId, s3Client, true))
```
davidsloan, line 45: For the source, this is always true?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3ConfigSettings.scala
```diff
@@ -138,4 +138,9 @@ object S3ConfigSettings {
   val SOURCE_ORDERING_TYPE:         String = s"$CONNECTOR_PREFIX.ordering.type"
   val SOURCE_ORDERING_TYPE_DOC:     String = "AlphaNumeric (the default)"
   val SOURCE_ORDERING_TYPE_DEFAULT: String = "AlphaNumeric"
+
+  val SOURCE_DELETE_MODE: String = s"$CONNECTOR_PREFIX.source.delete.mode"
+  val SOURCE_DELETE_MODE_DOC: String =
+    "Cleaning index files for GCP Cloud Storage via the compatible S3 APIs requires individual delete requests. Options are `BatchDelete` or `SeparateDelete`. Defaults to `BatchDelete`."
+  val SOURCE_DELETE_MODE_DEFAULT: String = "BatchDelete"
```
davidsloan, line 145: This property is labeled "SOURCE" but I thought this feature was for the sink?

davidsloan: Perhaps it could be a generic property for the connector, not related to SOURCE or SINK?

```
  val DELETE_MODE: String = s"$CONNECTOR_PREFIX.delete.mode"
```

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/state/S3SourceBuilder.scala
jdfoster: The Index Manager invokes the `deleteFiles` method and is only used in the sink functionality; therefore, the `batchDelete` parameter is redundant for the source pathway. I have set it to the default `true`, but it isn't needed for the source pathway. I could put the default value in the `AwsS3StorageInterface` constructor to `true` and not set the value here.

davidsloan (approved):

davidsloan:
Thanks for the contribution and the several rounds of changes.  Looks good so I have merged it in.

PR #958, opened 2023-06-13
RB-326

stheppi:
Handle scenarios where the partition discovery fails (i.e. connectivity issues). The partition discovery should not stop and resume.

The code changes PartitionDiscovery.run to handle errors on task:IO[Unit] ensuring the loop won't stop.

### DIFF
--- a/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/distribution/PartitionSearcher.scala
+++ b/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/distribution/PartitionSearcher.scala
@@ -26,14 +26,21 @@ import io.lenses.streamreactor.connect.aws.s3.storage.DirectoryFindCompletionCon
 import software.amazon.awssdk.services.s3.model.ListObjectsV2Request
 import software.amazon.awssdk.services.s3.model.ListObjectsV2Response
 
-class PartitionSearcher(
+trait PartitionSearcher {
+  def find(
+    lastFound: Seq[PartitionSearcherResponse],
+  ): IO[Seq[PartitionSearcherResponse]]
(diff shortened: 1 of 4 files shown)


kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/distribution/PartitionSearcher.scala
```diff
@@ -26,14 +26,21 @@ import io.lenses.streamreactor.connect.aws.s3.storage.DirectoryFindCompletionCon
...
 import software.amazon.awssdk.services.s3.model.ListObjectsV2Response
 
-class PartitionSearcher(
+trait PartitionSearcher {
+  def find(
+    lastFound: Seq[PartitionSearcherResponse],
+  ): IO[Seq[PartitionSearcherResponse]]
+}
+
+class PartitionSearcherImpl(
```
davidsloan, line 35: Can we find a better name than 'PartitionSearcherImpl'?  This seems a little too Java-esque to me.

Such as ScanningPartitionSearcher, AsyncPartitionSearcher, something that describes the specific implementation.

Great article [here](https://octoperf.com/blog/2016/10/27/impl-classes-are-evil/#impl-is-noise) I recommend reading.

davidsloan: Instead of introducing a trait, would it be better to pass in a function to `PartitionDiscovery` that performs the partition search?

`(Seq[PartitionSearcherResponse]) => IO[Seq[PartitionSearcherResponse]]`

davidsloan (approved):

davidsloan (approved):

PR #959, opened 2023-06-13
Control partitions discovery

stheppi:
Not all connectors need to have dynamic partition discovery. Some of the partitions are static. Running the partitions discovery once reduces the CPU and traffic to S3.

The code introduces a new property, of boolean type, to control the search. If set to true, the search is continuous, otherwise it is a one of.

PollLoop is extended with a new method which runs an IO until it is not failing. This is done to handle scenarios where there are intermittent failures (i.e. connectivity issues) and therefore the connector partition discovery can happen.


kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/source/S3SourceTaskTest.scala
```diff
@@ -100,7 +100,61 @@ class S3SourceTaskTest
           val props = DefaultProps
             .combine(
               Map(
-                "connect.s3.kcql" -> s"insert into ${bucketSetup.TopicName} select * from $BucketName:${bucketSetup.PrefixName}/$dir/ STOREAS `${format.entryName}$formatExtensionString` LIMIT 190",
+                KCQL_CONFIG -> s"insert into ${bucketSetup.TopicName} select * from $BucketName:${bucketSetup.PrefixName}/$dir/ STOREAS `${format.entryName}$formatExtensionString` LIMIT 190",
```
davidsloan, line 103: My only thought on replacing the values (eg "connect.s3.kcql") in the test with the constant, is it doesn't protect us if the constant value changes between versions, which is why it was originally implemented without referencing the constants.  But this is minor and is hopefully covered by other tests.

kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/source/S3SourceTaskTest.scala
```diff
@@ -100,7 +100,61 @@ class S3SourceTaskTest
...
+              ),
+            ).asJava
+
+          task.start(props)
+          //Let the partitions scan do its work
+          Thread.sleep(4000)
```
davidsloan, line 109: How would you feel about using `eventually` instead of adding artificial delays to the tests, which can introduce unreliability and slow the tests down unnecessarily?

I added a similar one to this yesterday, the example is here:
https://github.com/lensesio/stream-reactor/blob/cf3fca08f83043a59d2e73aa10b9e8c01451fba3/kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/source/S3SourceTaskXmlReaderTest.scala#L62

davidsloan (approved):

kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/source/S3SourceTaskTest.scala
stheppi: Eventually is good, but we know it will fail once or twice. I will adapt the tests. For me it does not give more in this case.

kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/source/S3SourceTaskTest.scala
stheppi: Sorry, I fail to fully understand. If we have the text there it protects from a change in KCQL_CONFIG? is that the benefit?

davidsloan: that was the gist of it yes

stheppi: Wouldn't a class with one test cover that? rather than use the actual string key and not the constant?

PR #961, opened 2023-06-15
337 - Enforce sink padding by default

davidsloan:
In order to ensure maximum compatibility with the S3 Source, set padding by default.

This will cause a change in behaviour to users upgrading.

**New Defaults**
* Default padding: LeftPad
* Default padding length: 12


stheppi:
As discussed, let's add more padding - maybe 12 0(zeros)?

stheppi (approved):

PR #966, opened 2023-07-11
Bring kafka-connect-query-language inside of Stream Reactor

davidsloan:
**KCQL Inside Stream Reactor**
(RB-312)
* Import commits from KCQL library into Stream Reactor to avoid having to build an external library.
* Configure sbt plugin.
* Move antlr files to antlr4 so the sbt plugin can detect these.
* Add headers and reformat.
* Remove gradle build files.
* Convert tests to scala.


build.sbt
```diff
@@ -44,7 +45,22 @@ lazy val root = (project in file("."))
   )
   .disablePlugins(AssemblyPlugin, HeaderPlugin)
 
+lazy val `query-language` = (project in file("kafka-connect-query-language"))
+  .settings(
+    settings ++
+      Seq(
+        name := "kafka-connect-query-language",
+        description := "Kafka Connect compatible connectors to move data between Kafka and popular data stores",
```
stheppi, line 53: nit-picking the project description might be better if it says it provides a SQL like DSL to configure the connectors behaviour

stheppi (approved):

build.sbt
davidsloan: Thanks  - will fix this in my next PR so as to avoid having to wait for the build again.

PR #984, opened 2023-09-08
Partition Name Keys & Naming Strategies Merge & Adding Extra Configuration Options

davidsloan:
- This PR represents a tidy-up of the naming strategies functionality and an evolution of the padding strategies.  This is to avoid so much duplication of code between the different naming strategies when in fact the difference is minimal.
- The Source Props Schema was previously being used for the Source + Sink although in most cases the properties should only target one or the other.  This has now been split out into `S3SourcePropsSchema` and `S3SinkPropsSchema`.
- There are major changes to configuration as a result of this PR and the default behaviour has been modified, so it is important that this information is well advertised to people using the sink.

### Implementation

**File Naming Strategies => S3FileNamer + S3KeyNamer**
S3KeyNamer and S3FileNamer replace the "FileNamingStrategy" code.

S3KeyNamer manages the partitions.  When using the default (hierarchical) partitioning strategy, this sets up the default partitions of "(Kafka) partition" and "offset".  This enables the same code to be reused, whereas previously there were implementations for each.

The S3KeyNamer is responsible for naming the key, and delegates to the S3FileNamer to name the file (the part of the key after the last `/`).

There are 2 implementations of S3FileNamer.

- OffsetS3FileNamer
  -  This is used with the default partitioning strategy and will produce a file with a name like `00000001.avro`.
- TopicPartitionOffsetS3FileNamer
  - This is used with the custom (partitioned) partitioning strategy and will produce a file with a name like `topic(1_0000001).avro`.

##### Naming / Partitioning

Mostly the configuration for naming and partitioning has not changed.  Please see the documentation.
[PARTITIONBY (partition[, partition] ...)]

However as an alternative for the `WITHPARTITIONER` KCQL
```
    WITHPARTITIONER=KeysAndValues
    WITHPARTITIONER=Values 
```

we have introduced the KCQL property `partition.include.keys`.  Remember only one of these should be set - and if the KCQL is set then this will take priority (until it is deprecated).

The WITHPARTITIONER now acts on both Default (Hierarchical) and Custom (Partitioned) modes - because now the difference between these modes is minimal, except they modify some configuration switches.

To retain legacy behaviour the key is shown by default for Custom partitioning, but only the value is shown for Default partitioning.

| Property | Title | Default (if not set) | Description |
|----------|-----|--------------------|------------|
| `partition.include.keys` | Partition Include Keys | false (for default) true (for custom)  | Includes keys in the partition name in the format key=value, eg partition=1 |

##### Padding

The old configuration for padding has been retained, however you should consider using the new configuration which is designed to be more configurable and easier to use.  If the key `connect.s3.padding.strategy` is set the legacy configuration method will be used.

The updated configuration for padding uses KCQL Properties.

The following KCQL properties have been added:

| Property | Title | Default (if not set) | Description |
|----------|-----|--------------------|------------|
| `padding.char` | Padding Character | '0' |
| `padding.length.offset` | Offset Padding Length | 12 |
| `padding.length.partition` | (Kafka) Partition Padding Length | 0 |
| `padding.type` | Padding Type | LeftPad | Type of padding to apply.  Can be LeftPad, RightPad or NoOp. |

#### Upgrading
Please see release notes when released.

#### New Setup
For a new setup the defaults are those we believe are the most sensible.


stheppi:
`  we have introduced the KCQL property partition.include.keys. Remember only one of these should be set - and if the KCQL is set then this will take priority (until it is deprecated).`

Just a placeholder to ask if we have code to check if both are set and we put a warning

stheppi:
`
padding.default.type
`
Would it read better if  `default` is removed from the property key?

stheppi:
We need a migration from 4.x to 5. on setting the padding values to keep them as they were before the upgrade

kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTaskJsonEnvelopeTest.scala
```diff
@@ -173,7 +173,7 @@ class S3SinkTaskJsonEnvelopeTest
     val props = DefaultProps
       .combine(
         Map(
-          "connect.s3.kcql" -> s"insert into $BucketName:$PrefixName select * from $TopicName STOREAS `JSON`  WITH_FLUSH_COUNT = 3 PROPERTIES('store.envelope'=true)",
+          "connect.s3.kcql" -> s"insert into $BucketName:$PrefixName select * from $TopicName STOREAS `JSON`  WITH_FLUSH_COUNT = 3 PROPERTIES('store.envelope'=true,'padding.fields.enable'='partition,offset')",
```
stheppi, line 176: it might be clear and less error-prone to have

padding.partition=true
padding.offset=true?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3ConfigSettings.scala
```diff
@@ -111,11 +111,11 @@ object S3ConfigSettings {
...
     "Configure in order to pad the partition and offset on the sink output files. Options are `LeftPad`, `RightPad` or `NoOp`  (does not add padding). Defaults to `LeftPad`."
-  val PADDING_STRATEGY_DEFAULT = "LeftPad"
+  val PADDING_STRATEGY_DEFAULT = ""
 
-  val PADDING_LENGTH         = s"$CONNECTOR_PREFIX.padding.length"
-  val PADDING_LENGTH_DOC     = s"Length to pad the string up to if $PADDING_STRATEGY is set."
-  val PADDING_LENGTH_DEFAULT = 12
+  val PADDING_LENGTH     = s"$CONNECTOR_PREFIX.padding.length"
+  val PADDING_LENGTH_DOC = s"Length to pad the string up to if $PADDING_STRATEGY is set."
+  val PADDING_LENGTH_DEFAULT: Int = -1
```
stheppi, line 118: Why is it -1? Did we not agree it will pad by default the file names? Also, -1 is a misleading value. If -1 is accepted, is -199?

davidsloan: This is the KCQL default.  This is the default value used in the config def.  If the KCQL is not set then it looks at the KCQL properties.  This happens in the code here:

https://github.com/lensesio/stream-reactor/blob/7d801c5229c69f62062a884957bc88b963ab674e/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/padding/PaddingService.scala#L39

Agree with your point on -199.  It is not a valid value.  In the code where I'm checking the default I will change it to a < comparison with zero so that it will consider **all** negatives as unset.

kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTaskJsonEnvelopeTest.scala
davidsloan: If we're going to do this, why not use it to set the padding length instead?  It seems like we should be able to have different fields with different padding lengths.

```
padding.partition=4
padding.offset=12
```

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/config/S3ConfigSettings.scala
davidsloan: Making the properties default value any other than -1 (a special value to show "unset") would be incompatible with having a check that indicates if both the properties value and KCQL properties are set and printing a warning.

kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTaskJsonEnvelopeTest.scala
stheppi: Indeed. Nice suggestion

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/TopicPartitionOffset.scala
```diff
@@ -24,6 +24,9 @@ case class Topic(value: String) extends AnyVal {
 
 object Offset {
 
+  // Define an implicit conversion from Long to Offset
+  implicit def longToOffset(value: Long): Offset = Offset(value)
```
stheppi, line 28: This is not advisable as a conversion. Offset itself it's a code syntactic sugar around Long. It's advisable to do offset(value) ad-hoc. The compiler, won't actually create a runtime instance of Offset

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/PartitionField.scala
```diff
@@ -97,6 +99,8 @@ case class TopicPartitionField() extends PartitionField {
 
 case class PartitionPartitionField() extends PartitionField {
```
stheppi, line 100: The naming is great :) partitionpartition :)

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/PartitionDisplay.scala
```diff
@@ -31,11 +37,21 @@ object PartitionDisplay extends Enum[PartitionDisplay] {
...
-        .withNameInsensitiveOption(_)
-        .getOrElse(KeysAndValues)
+  def apply(
+    kcql:    Kcql,
+    props:   KcqlProperties[S3PropsKeyEntry, S3PropsKeyEnum.type],
+    default: PartitionDisplay,
+  ): PartitionDisplay = fromProps(props).orElse(fromKcql(kcql)).getOrElse(default)
+
+  private def fromProps(props: KcqlProperties[S3PropsKeyEntry, S3PropsKeyEnum.type]): Option[PartitionDisplay] =
+    props.getOptionalBoolean(PartitionIncludeKeys) match {
```
stheppi, line 47: Option monad will take care of the None path
```
  props.getOptionalBoolean(PartitionIncludeKeys).map {
      case  true  => KeysAndValues 
      case  false => Values
    }
```

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/naming/S3KeyNamer.scala
```diff
@@ -35,115 +37,37 @@ import java.util.UUID
...
 }
-
-class PartitionedS3FileNamingStrategy(
+class S3KeyNamer(
   formatSelection:    FormatSelection,
-  paddingStrategy:    PaddingStrategy,
   partitionSelection: PartitionSelection,
-) extends S3FileNamingStrategy {
+  fileNamer:          S3FileNamer,
+  paddingFns:         Map[String, String => String],
```
stheppi, line 62: As discussed let's avoid the free form Map brings.

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/config/PartitionSearcherOptions.scala
```diff
@@ -17,8 +17,14 @@ package io.lenses.streamreactor.connect.aws.s3.source.config
...
+}
+
 case class PartitionSearcherOptions(
-  recurseLevels: Int,
-  continuous:    Boolean,
-  interval:      FiniteDuration,
+  recurseLevels:    Int,
+  continuous:       Boolean,
+  interval:         FiniteDuration,
+  wildcardExcludes: Set[String],
```
stheppi, line 29: Let's put a comment in the code on why this is needed.

kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3AvroWriterManagerTest.scala
```diff
@@ -63,11 +67,22 @@ class S3AvroWriterManagerTest extends AnyFlatSpec with Matchers with S3ProxyCont
...
+        commitPolicy    = CommitPolicy(Count(2)),
+        formatSelection = AvroFormatSelection,
+        keyNamer = new S3KeyNamer(
+          AvroFormatSelection,
+          defaultPartitionSelection(Values),
+          new HierarchicalS3FileNamer(
+            identity[String],
+            AvroFormatSelection.extension,
+          ),
+          Map[String, String => String](
```
stheppi, line 79: As discussed let's make use of the statically type compiler.

kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTaskJsonEnvelopeTest.scala
davidsloan: Thanks - I've implemented these as 

`padding.length.partition` and `padding.length.offset` (and implemented in a future-proof way that will not rule out allowing padding on custom fields)

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/TopicPartitionOffset.scala
davidsloan: OK, I got rid of the conversion and added an additional method

```
  def atOffset(offset: Long): TopicPartitionOffset = withOffset(Offset(offset))
```

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/PartitionField.scala
davidsloan: 🥔 🥔 🥔

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3DirectoryLister.scala
```diff
@@ -98,7 +109,11 @@ object AwsS3DirectoryLister extends LazyLogging {
                 if (levelsToRecurse > 0) {
                   acc + prefix
                 } else {
-                  if (connectorTaskId.ownsDir(prefix) && !exclude.contains(prefix)) acc + prefix
+                  if (
+                    connectorTaskId.ownsDir(prefix) && !exclude.contains(prefix) && !wildcardExclude.exists(we =>
+                      prefix.contains(we),
```
stheppi, line 121: isn't it enough to filter on level 0 and ignore the other parts? if we exclude the wildcard the code won't end up including keys under `.index/*`.

kafka-connect-aws-s3/src/test/scala/io/lenses/streamreactor/connect/aws/s3/sink/naming/S3KeyNamerTest.scala
```diff
@@ -75,6 +89,13 @@ class S3KeyNamerTest extends AnyFunSuite with Matchers with OptionValues with Ei
     UUID.fromString(uuid)
   }
 
+  test("finalFilename should write to the root of the bucket with no prefix") {
+
+    val result = s3KeyNamer.finalFilename(bucketNoPrefix, topicPartition, partitionValues)
+
+    result.value.path.value shouldEqual s"$TopicName/00$Partition/0$Offset.json"
```
stheppi, line 98: Isn't the padding set to 3?

> private val noOpPadder: String => String = LeftPadPaddingStrategy(3, '0').padString

Then why do we only get `00`

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/model/S3StoredFile.scala
```diff
@@ -1,53 +0,0 @@
-/*
```
stheppi, line 1: How was this replaced?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3WriterManager.scala
```diff
@@ -52,14 +53,14 @@ case class MapKey(topicPartition: TopicPartition, partitionValues: immutable.Map
...
 class S3WriterManager(
-  commitPolicyFn:       TopicPartition => Either[SinkError, CommitPolicy],
-  bucketAndPrefixFn:    TopicPartition => Either[SinkError, S3Location],
-  fileNamingStrategyFn: TopicPartition => Either[SinkError, S3FileNamingStrategy],
-  stagingFilenameFn:    (TopicPartition, Map[PartitionField, String]) => Either[SinkError, File],
-  finalFilenameFn:      (TopicPartition, Map[PartitionField, String], Offset) => Either[SinkError, S3Location],
-  formatWriterFn:       (TopicPartition, File) => Either[SinkError, S3FormatWriter],
-  indexManager:         IndexManager,
-  transformerF:         MessageDetail => Either[RuntimeException, MessageDetail],
+  commitPolicyFn:    TopicPartition => Either[SinkError, CommitPolicy],
```
stheppi, line 56: would it make sense to create an interface which can provide the info for : commitPolicyFn, bucketAndPRefixFn, and Key namer in one method .

The same with

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/reader/ResultReader.scala
```diff
@@ -41,17 +41,8 @@ class ResultReader(
...
-
-    if (results.isEmpty) {
-      None
-    } else {
-      Some(
-        results,
-      )
-    }
-
+    Option.when(results.nonEmpty)(results)
```
stheppi, line 45: Just a remark that this could also work `results.headOption.map(_=> results)`.

kafka-connect-aws-s3/src/test/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/PartitionDisplayTest.scala
```diff
@@ -0,0 +1,67 @@
...
+import io.lenses.streamreactor.connect.aws.s3.sink.config.PartitionDisplay.Values
+import io.lenses.streamreactor.connect.aws.s3.sink.config.kcqlprops.S3SinkPropsSchema
+import io.lenses.streamreactor.connect.config.kcqlprops.KcqlProperties
+import org.mockito.MockitoSugar
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class PartitionDisplayTest extends AnyFlatSpec with MockitoSugar with Matchers {
+
+  private val kcql: Kcql = mock[Kcql]
```
stheppi, line 32: The mock should be in every test. The best practice is for tests not to share any state.

stheppi (approved):

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/source/reader/ResultReader.scala
davidsloan: I think I prefer it the original way, taking the headOption and ignoring the value seems a little untidy

PR #985, opened 2023-09-08
S3 Sink: Allow backticks in KCQL partition fields

davidsloan:
To distinguish between flattened and nested json structures in configuration

Currently a dot is a special character in the `KCQL PARTITIONBY` statements (connector dependent)

This currently works well for nested structures.  For example:

`
"myObject": {
  "user": {
    name: "Bob"
  }
}
`

myObject.user.name will give us "Bob".

However it will not work for a flattened structure where the key contains a dot:
`
{ "myObject.user.name": "Bob" }
`
This PR ensures that you are able to specify in KCQL if the key should be respected whole by using backticks:

```
PARTITIONBY _key.`myObject.user.name`
```

where the part surrounded by backticks will be the section which must be used whole.

This fixes an issue reported via Lenses community Slack.

### DIFF
--- a/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/PartitionFieldSplitter.scala
+++ b/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/PartitionFieldSplitter.scala
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2017-2023 Lenses.io Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


stheppi (approved):

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/PartitionFieldSplitter.scala
```diff
@@ -0,0 +1,45 @@
...
+object PartitionFieldSplitter {
+
+  /**
+    * Splits an input string into parts while respecting backticks.
+    * Backticks are treated as complete strings and not subject to splitting.
+    *
+    * @param input The input string to split.
+    * @return A sequence of strings representing the split parts.
+    */
+  def split(input: String): Seq[String] =
```
stheppi, line 30:
```scala
    val input = "_value.field.`a.b`.c.d.`e.f.g`"
    val pattern  = """`([^`]+)`|([^.`]+)""".r

    
    val result = pattern.findAllMatchIn(input).map { matchResult =>
      Option(matchResult.group(1)).getOrElse(matchResult.group(2))
    }.toSeq
 
   result.foreach(println)
```

Achieves the same outcome. might be a bit easier to read (apart from the regex :) )

PR #986, opened 2023-09-16
fix(influxdb): Modify influxDbWriter to address the memory issue

OmalCooray:
During JVM profiling, I noticed the memory usage of the influxdb sink connector consistently increasing. Investigation revealed that the influxDbWriter wasn't closing properly, leading to the memory issue.


davidsloan (approved):

davidsloan:
Thank you for taking the time to investigate the problem and develop the fix!  We will bundle this into our upcoming release.

PR #995, opened 2023-10-25
Refactor S3 components into common module for reuse

davidsloan:
Remaining transformers into common.  Unit testing. Removing scala testcontainers lib from S3 and reusing the test-common S3 test containers Extract shared code to cloud-common
Renaming some common things which shouldn't have S3 in their name GCP Config WTD
Clean up, reduce tech debt, remove duplicated functionality, avoid weak string referencing Updating references to S3 in common module


FPerezP (approved):
From my limited knowledge of this code base, everything looks good. I would probably wait for Stefan to confirm but other than that, feel free to merge from my side. Thanks a lot for this refactoring and reorganization! Looks nice and necessary!

PR #999, opened 2023-11-08
Support BigDecimal

stheppi:
The PR introduces the support for S3 sink for BigDecimal.  A new case class extending SinkData is introduced. For text storage the safeValue function is overwritten to return the java BigDecimal. The JsonConverter is changed to store decimals as numbers, not base64 (default). While a breaking change, this should not be a cause of concern since BigDecimal was not handled before, so it should have created an error if that was the case.

Unit test and integration tests were added to ensure the storage can handle BigDecimal.


davidsloan (approved):
Looks good.

Similarly to this we should raise a jira to support java.util.Date in the future (translates to the logical types org.apache.kafka.connect.data.{Date/Time/Timestamp})

PR #1004, opened 2023-11-24
Feat/package standardisation

davidsloan:
Standardising the packages and configurations at "io.lenses", deprecating the former names "com.datamountaineer" and "com.landoop".

The documentation will be revamped in the next major release with warnings.

It would also be possible to put temporary classes in the old locations that print a warning the configuration needs to be updated and then delegate the the new locations.  I'm not sure this is necessary as this will be released in a major version change.


stheppi (approved):
let's capture the breaking changes upgrade boxes on the connector class name.z

PR #1008, opened 2023-11-29
Feat/5.0 changes

stheppi:
Brings 5.0 branch changes to master.

The S3 consumer offset sink will have another PR

### DIFF
--- a/kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3AvroWriterManagerTest.scala
+++ b/kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3AvroWriterManagerTest.scala
@@ -134,6 +134,69 @@ class S3AvroWriterManagerTest extends AnyFlatSpec with Matchers with S3ProxyCont
 
   }
 
+  "avro sink" should "write BigDecimal" in {
+    val sink = S3WriterManager.from(avroConfig)
+    val usersWithDecimal1 =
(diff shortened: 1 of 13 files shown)


kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3AvroWriterManagerTest.scala
```diff
@@ -134,6 +134,69 @@ class S3AvroWriterManagerTest extends AnyFlatSpec with Matchers with S3ProxyCont
...
 
+  "avro sink" should "write BigDecimal" in {
+    val sink = S3WriterManager.from(avroConfig)
+    val usersWithDecimal1 =
+      new Struct(UsersSchemaDecimal)
+        .put("name", "sam")
+        .put("title", "mr")
+        .put(
+          "salary",
+          BigDecimal(100.43).setScale(18).bigDecimal,
```
davidsloan, line 145: It's advisable to get into the habit of constructing using BigDecimal(String) instead of relying on floating-point literals. The conversion of floating-point literals, such as 100.43, to BigDecimal using setScale(18) may introduce rounding errors due to the binary representation of these literals. To ensure accurate representation and avoid precision loss, consider using BigDecimal("100.43").setScale(18) instead. 

Additionally, it's worth evaluating whether a scale of 18 is necessary for the specific use case, as an excessively high scale might impact performance and memory usage without providing significant benefits.


(This comment applies to various places in the PR.)

davidsloan (approved):

PR #1013, opened 2023-12-14
HTTP Sink MVP

davidsloan:
Work to date on a HTTP sink.

Not yet covered:
- deletion (via tombstones)
- tests with tls certificates

This is a work in progress and should not be used in a production environment.

**Configuration**
For speed of development the configuration is provided via a Kafka Connect configuration property (`connect.http.config`), through the means of a Json template.  Here is an example:

```json
{
   "authentication":{
      "username":"user",
      "password":"pass",
      "type":"BasicAuthentication"
   },
   "method":"Put",
   "endpoint":"http://{{clientSubDomain}}.example.com",
   "content":"<cost><costCentre>{{clientCostCentre}}</costCentre><spender>{{employee}}</spender><description>{{costDescription}}</description></body></cost>",
   "headers":[
      [
         "Content-Type",
         "application/xml"
      ],
      [
         "X-Client-Cost-Centre",
         "{{clientCostCentre}}"
      ]
   ]
}
```

There was a discussion about using KCQL however it was decided that this is not suitable for this use case.

**Templating**

Two different types of templates can be provided:

XML is used here as an example, in fact any markup language or text content can be supported.

***Single Template***

For when you want to substitute the contents of a single message into the template content.  It could be used for multiple messages but this wouldn't make much sense as there is no looping part.

```mustache
      <xml>
          <message>
            <topic>{{topic}}</topic>
            <employee>{{value.employeeId}}</employee>
            <order>{{value.orderNo}}</order>
            <groupDomain>{{value.groupDomain}}</groupDomain>
          </message>
      </xml>
```
***Multi Template***

For when you want to merge multiple messages inside a single HTTP request.  It loops through every message.
```mustache
      <xml>
        <messages>
          {{#message}}
          <message>
            <topic>{{topic}}</topic>
            <employee>{{value.employeeId}}</employee>
            <order>{{value.orderNo}}</order>
            <groupDomain>{{value.groupDomain}}</groupDomain>
          </message>
          {{/message}}
        </messages>
      </xml>
```


kafka-connect-http/src/main/scala/io/lenses/streamreactor/connect/http/sink/HttpSinkTask.scala
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  override def put(records: util.Collection[SinkRecord]): Unit = {
+    val sinkRecords = records.asScala
+    (maybeConfig, maybeTemplate) match {
+      case (Some(config), Some(template)) =>
+        val sender         = new HttpRequestSender(config)
+        val clientResource = JdkHttpClient.simple[IO]
+
+        clientResource.use {
```
cmcmteixeira, line 66: If worthwhile we can avoid allocating a client per put call.
A way that could be done would be to use the `.allocate` method on resource, something like what's done [here](https://github.com/lensesio-dev/lenses-core/blob/feat/RB-202_CE_3_Upgrade/flows-applications-http/src/test/scala/io/lenses/flows/application/util/DispatcherProvider.scala)

kafka-connect-http/src/main/scala/io/lenses/streamreactor/connect/http/sink/client/HttpRequestSender.scala
```diff
@@ -0,0 +1,58 @@
...
+
+    // Add authentication if present
+    val authenticatedRequest = config.authentication.fold(request) {
+      case BasicAuthentication(username, password) =>
+        request.putHeaders(Authorization(BasicCredentials(username, password)))
+    }
+
+    for {
+      response <- client.expect[String](authenticatedRequest)
+      _        <- IO(println(s"Response: $response"))
```
cmcmteixeira, line 53: There is a `IO.println` for a more idiomatic alternative

kafka-connect-http/src/main/scala/io/lenses/streamreactor/connect/http/sink/tpl/binding/KafkaConnectKeyBinding.scala
```diff
@@ -0,0 +1,27 @@
...
+package io.lenses.streamreactor.connect.http.sink.tpl.binding
+
+import cats.implicits.toBifunctorOps
+import io.lenses.streamreactor.connect.cloud.common.sink.extractors.KafkaConnectExtractor
+import org.apache.kafka.connect.errors.ConnectException
+import org.apache.kafka.connect.sink.SinkRecord
+
+class KafkaConnectKeyBinding(name: Option[String]) extends KafkaConnectBaseBinding() {
+  override def get(sinkRecord: SinkRecord): AnyRef = KafkaConnectExtractor.extractFromKey(sinkRecord, name).leftMap(e =>
+    throw new ConnectException(s"unable to extract field $name for template, ", e),
```
cmcmteixeira, line 25: It's a nit pick :) 
I would simply pattern match on the result of `extractFromKey` and throw on the left branch. It's a bit more explicit behavior wise

build.sbt
```diff
@@ -246,6 +246,28 @@ lazy val elastic7 = (project in file("kafka-connect-elastic7"))
   .configureFunctionalTests()
   .enablePlugins(PackPlugin)
 
+lazy val http = (project in file("kafka-connect-http"))
+  .dependsOn(common)
+  //.dependsOn(`test-common` % "fun->compile")
```
GoMati-MU, line 251: I guess the functional tests are to be added later? Do we need that line then?

build.sbt
```diff
@@ -246,6 +246,28 @@ lazy val elastic7 = (project in file("kafka-connect-elastic7"))
...
+        packExcludeJars := Seq(
+          "scala-.*\\.jar",
+          "zookeeper-.*\\.jar",
+        ),
+      ),
+  )
+  .configureAssembly(false)
+  .configureTests(baseTestDeps ++ kafkaConnectHttpTestDeps)
+  .configureIntegrationTests(baseTestDeps ++ kafkaConnectHttpTestDeps)
+  //.configureFunctionalTests(kafkaConnectS3FuncTestDeps)
```
GoMati-MU, line 268: I guess the functional tests are to be added later? Do we need that line then?

kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/commit/CloudCommitPolicy.scala
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+package io.lenses.streamreactor.connect.cloud.common.sink.commit
+
+import com.typesafe.scalalogging.LazyLogging
+import io.lenses.streamreactor.connect.cloud.common.sink.config.FlushSettings.defaultFlushCount
+import io.lenses.streamreactor.connect.cloud.common.sink.config.FlushSettings.defaultFlushInterval
+import io.lenses.streamreactor.connect.cloud.common.sink.config.FlushSettings.defaultFlushSize
+
+object CloudCommitPolicy extends LazyLogging {
```
cmcmteixeira, line 23: There's no need to extend LazyLogging here right ?

kafka-connect-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/commit/CommitContext.scala
```diff
@@ -15,20 +15,44 @@
...
   count:                Long,
   fileSize:             Long,
   createdTimestamp:     Long,
   lastFlushedTimestamp: Option[Long],
   partitionFile:        String,
-) {
-  def lastModified: Long = lastFlushedTimestamp.getOrElse(createdTimestamp)
+) extends CommitContext
+    with LazyLogging {
+  override def logFlush(flushing: Boolean, result: Seq[ConditionCommitResult]): Unit = {
```
cmcmteixeira, line 46: I wouldn't embed the logging call in the `CommitContext`. Generating the string would be fine but it feels to me that the representation of the commit context (data only) and a log line are two distinct concerns. 

One issue with embedding here the log line is that the line will say it was produced in the `CloudCommitContext` . It would likely be more useful it it mention that it was produced either in `HttpWriter` or `Writer`

cmcmteixeira: If it's not clear what I mean is that this method should only return the String and then it would be up to the Writer and/or HttpWriter to log it

kafka-connect-http/src/it/scala/io/lenses/streamreactor/connect/http/sink/HttpSinkTaskIT.scala
```diff
@@ -0,0 +1,325 @@
...
+    wireMockServer.stop()
+
+  object HttpSinkTaskReleaseable extends Releasable[HttpSinkTask] {
+    override def release(resource: HttpSinkTask): Unit = {
+      logger.info("SHUTTING DOWN SINK TASK")
+      resource.stop()
+    }
+  }
+
+  implicit val releaseSinkTask: Releasable[HttpSinkTask] = HttpSinkTaskReleaseable
```
cmcmteixeira, line 56: I'm would standardize on a single way to manage Resource acquisition/release. In this test we kind of have two levels of it. One managed by the "BeforeAndAfterAll" scalatest capability and another by scala's `Using`. 

Instead of doing this I would suggest standardizing on the cats effect Resource.

I've tried it locally and I've came up with the code bellow. However I'm having two issues 1) wiremock seems to be thread aware somehow and 2) the release of the http sink task is not actually stopping it from running. 

I'm having a look as to why :eye:  
```scala
package io.lenses.streamreactor.connect.http.sink

import cats.effect.IO
import cats.effect.Resource
import cats.effect.testing.scalatest.AsyncIOSpec
import cats.implicits.catsSyntaxOptionId
import cats.implicits.none
import com.github.tomakehurst.wiremock.WireMockServer
import com.github.tomakehurst.wiremock.client.WireMock.{ post => httpPost }
import com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo
import com.github.tomakehurst.wiremock.client.WireMock.aResponse
import com.github.tomakehurst.wiremock.client.WireMock.exactly
import com.github.tomakehurst.wiremock.client.WireMock.urlMatching
import com.github.tomakehurst.wiremock.client.WireMock.containing
import com.github.tomakehurst.wiremock.client.WireMock.equalToXml
import com.github.tomakehurst.wiremock.client.WireMock.postRequestedFor
import com.github.tomakehurst.wiremock.core.WireMockConfiguration.wireMockConfig
import io.lenses.streamreactor.connect.http.sink.client.HttpMethod
import io.lenses.streamreactor.connect.http.sink.config.BatchConfiguration
import io.lenses.streamreactor.connect.http.sink.config.HttpSinkConfig
import org.apache.kafka.connect.errors.ConnectException
import org.apache.kafka.connect.sink.SinkRecord
import org.scalatest.concurrent.Eventually
import org.scalatest.funsuite.AsyncFunSuite
import scala.jdk.CollectionConverters.MapHasAsJava
import scala.jdk.CollectionConverters.SeqHasAsJava

class HttpSinkTaskIT extends AsyncFunSuite with AsyncIOSpec with Eventually {

  def wireMockServer: Resource[IO, WireMockServer] =
    for {
      server <- Resource.eval(IO.delay(new WireMockServer(wireMockConfig().dynamicPort())))
      _      <- Resource.make(IO.delay(server.start()))(_ => IO.delay(server.stop()))
    } yield server

  def sinkTask(config: String): Resource[IO, HttpSinkTask] =
    for {
      task <- Resource.eval(IO.delay(new HttpSinkTask()))
      _    <- Resource.make(IO.delay(task.start(Map("connect.http.config" -> config).asJava)))(_ => IO.delay(task.stop()))
    } yield task

  private val Host  = "localhost"
  private val users = SampleData.Employees

  test("data triggers post calls") {
    val path = "/awesome/endpoint"
    (for {
      server <- wireMockServer
      configuration = HttpSinkConfig(
        authentication   = Option.empty,
        method           = HttpMethod.Post,
        endpoint         = s"http://$Host:${server.port()}/awesome/endpoint",
        content          = "test",
        headers          = Seq(),
        sslConfig        = Option.empty,
        batch            = BatchConfiguration(2L.some, none, none).some,
        errorThreshold   = none,
        uploadSyncPeriod = none,
      ).toJson
      _         = server.stubFor(httpPost(urlEqualTo(path)).willReturn(aResponse().withStatus(200)))
      sinkTask <- sinkTask(configuration)
      _ = sinkTask.put(
        users.zipWithIndex.map {
          case (struct, i) => new SinkRecord("myTopic", 0, null, null, SampleData.EmployeesSchema, struct, i.toLong)
        }.asJava,
      )
    } yield server).use { server =>
      IO.delay(eventually(server.verify(exactly(3), postRequestedFor(urlEqualTo(path)))))
    }
  }

  test("data triggers post calls to individual templated endpoints for single records") {
    val path = "/awesome/endpoint/.*"
    (for {
      server <- wireMockServer
      config = HttpSinkConfig(
        authentication   = Option.empty,
        method           = HttpMethod.Post,
        endpoint         = s"http://$Host:${server.port()}/awesome/endpoint/{{value.name}}",
        content          = "{salary: {{value.salary}}}",
        headers          = Seq(),
        sslConfig        = Option.empty,
        batch            = BatchConfiguration(1L.some, none, none).some,
        errorThreshold   = none,
        uploadSyncPeriod = none,
      ).toJson
      _         = server.stubFor(httpPost(urlMatching(path)).willReturn(aResponse().withStatus(200)))
      sinkTask <- sinkTask(config)
      _ = sinkTask.put(
        users.zipWithIndex.map {
          case (struct, i) => new SinkRecord("myTopic", 0, null, null, SampleData.EmployeesSchema, struct, i.toLong)
        }.asJava,
      )
    } yield server).use { server =>
      // verify REST calls
      IO.delay {
        eventually {
          server.verify(
            exactly(1),
            postRequestedFor(urlEqualTo("/awesome/endpoint/martin")).withRequestBody(containing("{salary: 35896.00}")),
          )
          server.verify(exactly(1), postRequestedFor(urlEqualTo("/awesome/endpoint/jackie")))
          server.verify(exactly(1), postRequestedFor(urlEqualTo("/awesome/endpoint/adam")))
          server.verify(exactly(1), postRequestedFor(urlEqualTo("/awesome/endpoint/jonny")))
          server.verify(exactly(1), postRequestedFor(urlEqualTo("/awesome/endpoint/jim")))
          server.verify(exactly(1), postRequestedFor(urlEqualTo("/awesome/endpoint/wilson")))
          server.verify(exactly(1), postRequestedFor(urlEqualTo("/awesome/endpoint/milson")))
        }
      }
    }
  }

  // TODO: I don't think this is a valid use case unless you want to aggregate records.  It doesn't really make sense, perhaps this should throw an error instead.
  test("data batched to single endpoint for multiple records using a simple template uses the first record") {
    val path = "/awesome/endpoint/.*"
    (for {
      server <- wireMockServer
      configuration = HttpSinkConfig(
        authentication   = Option.empty,
        method           = HttpMethod.Post,
        endpoint         = s"http://$Host:${server.port()}/awesome/endpoint/{{value.name}}",
        content          = "{salary: {{value.salary}}}",
        headers          = Seq(),
        sslConfig        = Option.empty,
        batch            = BatchConfiguration(7L.some, none, none).some,
        errorThreshold   = none,
        uploadSyncPeriod = none,
      ).toJson
      _     = server.stubFor(httpPost(urlMatching(path)).willReturn(aResponse().withStatus(200)))
      task <- sinkTask(configuration)
      _ = task.put(
        users.zipWithIndex.map {
          case (struct, i) => new SinkRecord("myTopic", 0, null, null, SampleData.EmployeesSchema, struct, i.toLong)
        }.asJava,
      )
    } yield server).use { server =>
      IO.delay {
        eventually {
          server.verify(
            exactly(1),
            postRequestedFor(urlEqualTo("/awesome/endpoint/martin")).withRequestBody(containing("{salary: 35896.00}")),
          )
        }
      }
    }

  }

  test("data batched to single endpoint for multiple records using a loop template") {
    val path = "/awesome/endpoint/.*"
    val expected =
      """
        |<salaries>
        |  <salary>35896.00</salary>
        |  <salary>60039.00</salary>
        |  <salary>65281.00</salary>
        |  <salary>66560.00</salary>
        |  <salary>63530.00</salary>
        |  <salary>23309.00</salary>
        |  <salary>10012.00</salary>
        |</salaries>
        |""".stripMargin

    (for {
      server <- wireMockServer
      configuration = HttpSinkConfig(
        authentication = Option.empty,
        method         = HttpMethod.Post,
        endpoint       = s"http://$Host:${server.port()}/awesome/endpoint/{{value.name}}",
        content =
          s"""
             | <salaries>
             |  {{#message}}
             |    <salary>{{value.salary}}</salary>
             |   {{/message}}
             | </salaries>""".stripMargin,
        headers          = Seq(),
        sslConfig        = Option.empty,
        batch            = BatchConfiguration(7L.some, none, none).some,
        errorThreshold   = none,
        uploadSyncPeriod = none,
      ).toJson
      _     = server.stubFor(httpPost(urlMatching(path)).willReturn(aResponse().withStatus(200)))
      task <- sinkTask(configuration)
      _ = task.put(
        users.zipWithIndex.map {
          case (struct, i) => new SinkRecord("myTopic", 0, null, null, SampleData.EmployeesSchema, struct, i.toLong)
        }.asJava,
      )
    } yield server).use { server =>
      IO.delay {
        // verify REST calls
        eventually {
          server.verify(
            exactly(1),
            postRequestedFor(urlEqualTo("/awesome/endpoint/martin")).withRequestBody(equalToXml(expected)),
          )
        }
      }
    }
  }

  test("broken endpoint will return failure and error will be thrown") {

    val path = "/awesome/endpoint"
    (for {
      server <- wireMockServer
      configuration = HttpSinkConfig(
        Option.empty,
        HttpMethod.Post,
        s"http://$Host:${server.port()}/awesome/endpoint",
        s"""
           | Ultimately not important for this test""".stripMargin,
        Seq(),
        Option.empty,
        BatchConfiguration(
          1L.some,
          none,
          none,
        ).some,
        none,
        none,
      ).toJson
      _     = server.stubFor(httpPost(urlMatching(path)).willReturn(aResponse().withStatus(404)))
      task <- sinkTask(configuration)
    } yield task).use { task =>
      IO.delay {
        eventually {
          // put data
          assertThrows[ConnectException] {
            task.put(
              users.zipWithIndex.map {
                case (struct, i) =>
                  new SinkRecord("myTopic", 0, null, null, SampleData.EmployeesSchema, struct, i.toLong)
              }.asJava,
            )
          }
        }
      }
    }
  }

}
```

kafka-connect-http/src/main/scala/io/lenses/streamreactor/connect/http/sink/HttpSinkTask.scala
```diff
@@ -0,0 +1,191 @@
...
+    printAsciiHeader(manifest, "/http-sink-ascii.txt")
+
+    val propsAsScala = props.asScala
+    maybeSinkName = propsAsScala.get("name")
+
+    IO.fromEither(parseConfig(propsAsScala.get(configProp)))
+      .flatMap { config =>
+        val template      = RawTemplate(config.endpoint, config.content, config.headers)
+        val writerManager = HttpWriterManager(sinkName, config, template)
+
```
cmcmteixeira, line 66: I believe you missed setting the `maybeWriterManager` here

kafka-connect-http/src/main/scala/io/lenses/streamreactor/connect/http/sink/HttpWriterManager.scala
```diff
@@ -0,0 +1,202 @@
...
+    for {
+      _ <- IO(logger.info(s"[$sinkName] starting HttpWriterManager"))
+      runnable: Runnable = () => {
+        process().flatMap {
+          resultList =>
+            handleResult(resultList, errCallback)
+        }.unsafeRunSync()
+        ()
+      }
+      _ <- IO.delay(scheduler.scheduleAtFixedRate(runnable, 0, uploadSyncPeriod.toLong, TimeUnit.SECONDS))
```
cmcmteixeira, line 153: Let's have a look at this on Monday maybe. 

Unfortunately, doing it like this means that it will be ever running and cannot be stopped. 
Running the test from `HttpSinkTaskIT` you can see that even tho the test terminates this process keeps running until the jvm is terminated. Ideally we would want to have a "stop" method that could interrupt this scheduling.

kafka-connect-http/src/main/scala/io/lenses/streamreactor/connect/http/sink/HttpSinkTask.scala
```diff
@@ -0,0 +1,191 @@
...
+import java.util
+import scala.jdk.CollectionConverters.IterableHasAsScala
+import scala.jdk.CollectionConverters.MapHasAsJava
+import scala.jdk.CollectionConverters.MapHasAsScala
+
+class HttpSinkTask extends SinkTask with LazyLogging {
+  private val manifest = JarManifest(getClass.getProtectionDomain.getCodeSource.getLocation)
+
+  implicit val ec: scala.concurrent.ExecutionContext = scala.concurrent.ExecutionContext.global
```
cmcmteixeira, line 46: this is unnused. 

Also, to make things more obvious using `implicit val runtime = IORuntime.global` rather than using the import is a bit better

kafka-connect-http/src/it/scala/io/lenses/streamreactor/connect/http/sink/HttpSinkTaskIT.scala
cmcmteixeira: (I've edited the above.. it was missing the patience config so it was failing)

kafka-connect-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/commit/CommitContext.scala
davidsloan: I definitely understand what you mean.  This would be better.

The reason it is done like this, is to make the most of the typesafe logger framework's macro support.

Simply returning a string, If debug is disabled, then all the evaluation and string concatenation etc will occur but will be wasted.

I wonder if I could follow your approach but instead of returning a String, return a () => String so as to avoid the unnecessary processing?

kafka-connect-http/src/main/scala/io/lenses/streamreactor/connect/http/sink/HttpSinkTask.scala
davidsloan: I think if you are missing the implicit, this is what IntelliJ suggests you add.  Will have to remember not to go with the default suggestion!

kafka-connect-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/commit/CommitContext.scala
cmcmteixeira: Indeed,  good point ! 

Having a look at the macro itself however I believe it might actually be ok, provided you don't instantiate the string outside the log call.

```scala
 q"if ($underlying.isDebugEnabled) $underlying.debug($message, _root_.scala.Array[AnyRef](${anyRefArgs.head}, ${anyRefArgs(1)}): _*)"
 ```
 
 Seems like it just rewrites the line as: 
 ```scala
 if(logger.isDebugEnabled){
    logger.debug(...)
 }
 ```
 
 so the string won't be allocated unless debug is enabled

kafka-connect-http/src/main/scala/io/lenses/streamreactor/connect/http/sink/HttpSinkTask.scala
cmcmteixeira: It's not a major problem.. It just highlights the fact that we're "using" the IORuntime here.. 

Imports are often overlooked and or not read so it makes it a bit harder to understand where the runtime instance is coming from(I may have been spending too much time tracking this things lately so that's why I'm a bit sensitive :wink: )

cmcmteixeira (approved):

PR #1025, opened 2024-01-31
Byte Array Support

stheppi:
Storing the Key/Value as an array of bytes has an issue since the connector returns java.nio.ByteBuffer but the Connect framework ByteArrayConverter works only with byte[]

This change ensures that if the key/value are a ByteBuffer it converts that to byte[]

### DIFF
--- a/kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/formats/reader/converters/SchemaAndValueEnvelopeConverter.scala
+++ b/kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/formats/reader/converters/SchemaAndValueEnvelopeConverter.scala
@@ -25,6 +25,7 @@ import org.apache.kafka.connect.header.ConnectHeaders
 import org.apache.kafka.connect.header.Headers
 import org.apache.kafka.connect.source.SourceRecord
 
+import java.nio.ByteBuffer
 import java.time.Instant
 import scala.jdk.CollectionConverters.ListHasAsScala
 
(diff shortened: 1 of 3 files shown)


stheppi:
∑π´≠œasz

kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/formats/reader/converters/SchemaAndValueEnvelopeConverter.scala
```diff
@@ -80,13 +81,15 @@ class SchemaAndValueEnvelopeConverter(
     if (fields.contains("key")) {
       key       = struct.get("key")
       keySchema = struct.schema().field("key").schema()
+      key       = byteBufferToArray(keySchema, key)
     }
```
davidsloan, line 85: Do you think it would be better if we could avoid the mutable var state and also overwriting the same value immediately.

kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/formats/reader/converters/SchemaAndValueEnvelopeConverter.scala
```diff
@@ -80,13 +81,15 @@ class SchemaAndValueEnvelopeConverter(
...
 
     var value:       Any    = null
     var valueSchema: Schema = null
     if (fields.contains("value")) {
       value       = struct.get("value")
       valueSchema = struct.schema().field("value").schema()
+      value       = byteBufferToArray(valueSchema, value)
     }
```
davidsloan, line 93: Same comment as above but will raise you a "this code is identical*, can it be refactored into a reusable function"?

* with the exception of some variable names

PR #1029, opened 2024-02-20
LC-68 - Avoid passing round java maps

davidsloan:
Avoid passing round java maps.  Only on ingress and egress from the scala code

This has caused a maintainability issue with the main and test code and is making refactors difficult due to the number of conversions required.

The SinkTask and SourceTask are bound to the Java API, and when we manage connect ConfigDef we are also bound to the Java API.  However these are at very distinct points in the code.  Currently we have conversions from/to java maps happening in a haphazard way throughout the codebase and this is causing maintainability issues and preventing sensible refactoring.  Enough is enough and this PR attempts to fix this so we can move forward.

### DIFF
--- a/kafka-connect-aws-s3/src/test/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/S3SinkConfigDefBuilderTest.scala
+++ b/kafka-connect-aws-s3/src/test/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/S3SinkConfigDefBuilderTest.scala
@@ -33,7 +33,7 @@ import org.scalatest.matchers.should.Matchers
 import scala.concurrent.duration.DurationInt
 import scala.jdk.CollectionConverters.IteratorHasAsScala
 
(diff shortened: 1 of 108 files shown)


kafka-connect-aws-s3/src/test/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/S3SinkConfigDefBuilderTest.scala
```diff
@@ -32,9 +32,8 @@ import org.scalatest.matchers.should.Matchers
 
 import scala.concurrent.duration.DurationInt
 import scala.jdk.CollectionConverters.IteratorHasAsScala
-import scala.jdk.CollectionConverters.MapHasAsJava
 
-class S3S3S3SinkConfigDefBuilderTest extends AnyFlatSpec with MockitoSugar with Matchers with EitherValues {
+class S3SinkConfigDefBuilderTest extends AnyFlatSpec with MockitoSugar with Matchers with EitherValues {
```
davidsloan, line 36: I have no idea what happened here 😁

stheppi (approved):

PR #1033, opened 2024-02-28
Azure EventHubs Connector MVP

GoMati-MU:
Adds Azure EventHubs Connector MVP. Implementation based on Kafka API being implemented by EventHubs, it doesn't (yet) leverage neither EventHubs native API nor KCQL (with latter being at this moment considered not needed)


java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/config/AzureEventHubsConfig.java
```diff
@@ -1,154 +1,185 @@
...
+    configDefinition = new ConfigDef(kafkaConsumerConfigToExpose)
+        .define(AzureEventHubsConfigConstants.EVENTHUB_NAME,
+            Type.STRING,
+            Importance.HIGH,
+            AzureEventHubsConfigConstants.EVENTHUB_NAME_DOC,
+            CONNECTION_GROUP,
+            1,
+            ConfigDef.Width.LONG,
+            AzureEventHubsConfigConstants.EVENTHUB_NAME
+            //CONFIGS BELOW IDENTIFIED AS NOT NEEDED (for now)
```
stheppi, line 42: let's remove the commented-out code.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/config/AzureEventHubsConfig.java
```diff
@@ -1,154 +1,185 @@
...
+        ).define(AzureEventHubsConfigConstants.KCQL_CONFIG,
+            Type.STRING,
+            AzureEventHubsConfigConstants.KCQL_DEFAULT,
+            Importance.HIGH,
+            AzureEventHubsConfigConstants.KCQL_DOC,
+            "Mappings",
+            1,
+            ConfigDef.Width.LONG,
+            AzureEventHubsConfigConstants.KCQL_CONFIG
+        ).define(AzureEventHubsConfigConstants.INCLUDE_HEADERS,
```
stheppi, line 133: I would leave this out until is requested as a feature, and always include headers.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/config/AzureEventHubsConfig.java
```diff
@@ -1,154 +1,185 @@
...
+            Type.BOOLEAN,
+            AzureEventHubsConfigConstants.INCLUDE_HEADERS_DEFAULT,
+            Importance.MEDIUM,
+            AzureEventHubsConfigConstants.INCLUDE_HEADERS_DOC,
+            "Mappings",
+            2,
+            ConfigDef.Width.LONG,
+            AzureEventHubsConfigConstants.INCLUDE_HEADERS
+        ).define(
```
stheppi, line 142: The nbr retries is from the old approach for handling failure and have the retry in the connector. With Connect framework improved retries , this option is not required.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureConsumerRebalancerListener.java
```diff
@@ -0,0 +1,59 @@
...
+    partitions.forEach(partition -> {
+
+      if (topicPartitionOffsetProvider.isPresent()) {
+        TopicPartitionOffsetProvider offsetProvider = topicPartitionOffsetProvider.get();
+        AzureTopicPartitionKey partitionKey = new AzureTopicPartitionKey(
+            partition.topic(), partition.partition());
+        Optional<AzureOffsetMarker> partitionOffset = offsetProvider.getOffset(partitionKey);
+        partitionOffset.ifPresentOrElse(
+            offset -> kafkaConsumer.seek(partition, offset.getOffsetValue()),
+            () -> kafkaConsumer.seekToBeginning(Collections.singletonList(partition)));
```
stheppi, line 51: Let's track a note to allow this option between earliest and latest to be configurable.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/TopicPartitionOffsetProvider.java
```diff
@@ -0,0 +1,113 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.apache.kafka.connect.storage.OffsetStorageReader;
+
+/**
+ * This class represents an abstraction over OffsetStorageReader that can be freely called by Azure
+ * EventHub Kafka Consumers when it was initialized once. It helps consumers to find out which
+ * offset were already committed inside Kafka Connect.
+ */
+@Slf4j
+public final class TopicPartitionOffsetProvider {
```
stheppi, line 15: Why is there a need to have a static instance? 
Can getOffset method not be a static method which also receives OffsetStorageReader as parameter?

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureEventHubsSourceTask.java
```diff
@@ -28,29 +45,51 @@ public String version() {
 
   @Override
   public void start(Map<String, String> props) {
-    log.info("Task initialising");
-    printAsciiHeader(jarManifest, "/azure-eventhubs-ascii.txt");
+    Optional<SourceTaskContext> contextOptional = ofNullable(context);
+    Map<String, String> configProps = null;
+    if (contextOptional.isPresent()) {
```
stheppi, line 50: It's down to my lack of knowledge maybe, but why is the `context` here used to get the properties and ignore the `props` parameter provided here?

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureEventHubsSourceTask.java
```diff
@@ -28,29 +45,51 @@ public String version() {
...
+      configProps = context.configs();
+      ofNullable(context.offsetStorageReader())
+          .ifPresent(TopicPartitionOffsetProvider::initialize);
```
stheppi, line 53: As mentioned, i think this does not add value creating an instance, deal with synchronisation just to have the reference on offsetStorageReader.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureEventHubsSourceTask.java
```diff
@@ -28,29 +45,51 @@ public String version() {
...
+    if (configProps == null || configProps.isEmpty()) {
       configProps = props;
-    } else {
-      assert context != null;
-      configProps = context.configs();
     }
+
+    eventHubsKafkaConsumerController = new EventHubsKafkaConsumerController(
+        new AzureEventHubsConfig(configProps), new KafkaConsumerProvider(),
+        new ArrayBlockingQueue<>(BLOCKING_QUEUE_DEFAULT_CAPACITY));
```
stheppi, line 62: Is there a Jira to track this? The 20-items capacity is not a production-ready default. Both frameworks, depending on the message size can handle a lot more records per second. This option should also be configurable, or probably be a factor of max.poll.records settings. If the client can poll 1000 records in a poll, maybe the queue should be N x max.poll.records

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureEventHubsSourceTask.java
```diff
@@ -28,29 +45,51 @@ public String version() {
...
+   * {@link EventHubsKafkaConsumerController} instance.
+   *
+   * @param eventHubsKafkaConsumerController {@link EventHubsKafkaConsumerController} for this task
+   * @param topic topic to subscribe to
+   */
+  public void initialize(EventHubsKafkaConsumerController eventHubsKafkaConsumerController,
+      String topic) {
+    this.eventHubsKafkaConsumerController = eventHubsKafkaConsumerController;
+    log.info("{} initialising.", getClass().getSimpleName());
+    printAsciiHeader(jarManifest, "/azure-eventhubs-ascii.txt");
```
stheppi, line 78: I am not sure if the ascii should be present in the task. 100 tasks, then there will be 100 ascii. Maybe we keep it just in the connector.

@davidsloan ?

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/BlockingQueuedKafkaConsumer.java
```diff
@@ -0,0 +1,81 @@
...
+  private class EventhubsPollingRunnable implements Runnable {
+    private final Duration pollDuration;
+
+    public EventhubsPollingRunnable(Duration pollDuration) {
+      this.pollDuration = pollDuration;
+    }
+
+    @Override
+    public void run() {
+      while (true) {
```
stheppi, line 65: When a task is stopped this thread needs to be stopped. Having this while loop means leaking this thread and that could easily end up consuming high CPU.

This thread lifespan should be correlated with the connector task lifespan.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/BlockingQueuedKafkaConsumer.java
```diff
@@ -0,0 +1,81 @@
...
+        ConsumerRecords<String, String> consumerRecords = consumer.poll(pollDuration);
```
stheppi, line 66: What happens if `poll` fails?

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
```diff
@@ -0,0 +1,144 @@
...
+
+    queuedKafkaConsumer = createQueuedConsumer();
+  }
+
+  private BlockingQueuedKafkaConsumer createQueuedConsumer() {
+    Properties consumerProperties = extractConsumerProperties();
+    return kafkaConsumerProvider.createConsumer(consumerProperties, recordsQueue);
+  }
+
+  private Properties extractConsumerProperties() {
```
stheppi, line 64: Why not extract the Kafka Consumer client initialisation?

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
```diff
@@ -0,0 +1,144 @@
...
+    Properties consumerProperties = new Properties();
+
+    consumerProperties.put(BOOTSTRAP_SERVERS_CONFIG,
```
stheppi, line 67: I am confused about this method. The connector configuration has `....connection.settings`. All that is required is to extract those properties and their value and overwrite:

* group id, to use the connector name
* deserializer which depend on the incoming data format

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
```diff
@@ -0,0 +1,144 @@
...
+    ConsumerRecords<String, String> consumerRecords = null;
+    try {
+      consumerRecords = recordsQueue.poll(
+          duration.get(ChronoUnit.SECONDS), TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
+      log.info("{} has been interrupted on poll", this.getClass().getSimpleName());
+      throw e;
+    }
+
+    if (consumerRecords != null && !consumerRecords.isEmpty()) {
```
stheppi, line 117: The null and empty check can be avoided if there is a decision where the code never returns null. The the logic of the method can be reduced to:

```
if(!empty){
return consumerRecords.stream.map{
  convert to source record
} .collect(Collectors.toList).
} else{
return Collections.emptyList();
}
```

This also avoids checking on sourceRecord being initialised.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
```diff
@@ -0,0 +1,144 @@
...
+            mapSourceRecordWithoutHeaders(consumerRecord, azureTopicPartitionKey,
+                offsetMarker, Schema.OPTIONAL_STRING_SCHEMA, Schema.STRING_SCHEMA);
+
+        sourceRecords.add(sourceRecord);
+      }
+    }
+    return sourceRecords != null ? sourceRecords : Collections.emptyList();
+  }
+
+  void subscribeToTopic(String topic) {
```
stheppi, line 137: I think this indirection does not help. I am repeating myself here, but the consumer and its subscription can be created at once since everything is driven by config which does not change for the lifspam of a connector task. 

In this case the task calls the controller to subscribe, the controller calls the blockingconsumer to subscribe. Testing the code becomes tedious.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/KafkaConsumerProvider.java
```diff
@@ -0,0 +1,38 @@
...
+   *
+   * @param consumerProperties Properties for Kafka Consumer
+   * @param recordBlockingQueue BlockingQueue for ConsumerRecords
+   * @return BlockingQueuedKafkaConsumer instance.
+   */
+  public BlockingQueuedKafkaConsumer createConsumer(Properties consumerProperties,
+      BlockingQueue<ConsumerRecords<String, String>> recordBlockingQueue) {
+    String clientId;
+    synchronized (EventHubsKafkaConsumerController.class) {
+      clientId = "KafkaEventHubConsumer#" + id++;
```
stheppi, line 29: let's consider two connector tasks running in parallel. 
Each one will run this code and for each the `id` starts at zero and thus create the same client id. The static state is not shared between tasks. If you want to control the id, you can control from the Connector:start by enriching the config to add the order/sequence.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureEventHubsSourceTask.java
GoMati-MU: I moved it to Connector for the time being, I can move it back if we feel like it.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/config/AzureEventHubsConfig.java
GoMati-MU: Happy to do so! Didn't want to remove them straight away as they were present in original story (LC-107)

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/config/AzureEventHubsConfig.java
GoMati-MU: It defaults to `true` do it includes headers by default. Shall we remove it anyway even though it's a working code?

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/config/AzureEventHubsConfig.java
GoMati-MU: Removed, thanks!

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureConsumerRebalancerListener.java
GoMati-MU: Shall I make a JIRA about it or add and expose a config param now?

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/TopicPartitionOffsetProvider.java
GoMati-MU: My logic was: this component is Connector-wide which means it's a good candidate for app-wide Singleton.

We can make it non-singleton but then big downside will be that I will initialize it in Connector class (because it has `OffsetStorageReader`) and pass it to every class (whether it uses it or not) so it'll eventually be passed to KafkaConsumers. IMO it's not worth it, but feel free to suggest another approach.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureEventHubsSourceTask.java
GoMati-MU: Let's keep in mind this is not 20 records: this is 20 PACKS of records (`ConsumerRecords`) class, not single `ConsumerRecord`). But I think it makes sense for it to be configurable, I added this.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureEventHubsSourceTask.java
GoMati-MU: I moved it to Connector for the time being, I can move it back if we feel like it.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/BlockingQueuedKafkaConsumer.java
```diff
@@ -0,0 +1,81 @@
...
+  void startPolling(Duration pollDuration) {
+    if (!isPolling) {
+      Thread pollingThread = new Thread(new EventhubsPollingRunnable(pollDuration));
+
+      pollingThread.start();
+      isPolling = true;
+    }
+  }
+
+  public void subscribe(List<String> strings) {
```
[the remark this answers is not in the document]
GoMati-MU: Good points! I amended the class to:
- taken into account that it's a Producer (it's a Consumer too on the inside, but looking from outside it's definitely a producer)
- Fitted topic subscribing into initialization, there's no point to call it from Connector/Task
- Changed methods to stop/start logic
- Added AtomicBoolean around volatility

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/BlockingQueuedKafkaConsumer.java
GoMati-MU: Good point! Amended implementation

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/BlockingQueuedKafkaConsumer.java
GoMati-MU: As for now the Runtime exception will be bubbled up: I believe that's what we wanted for MVP?

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
```diff
@@ -0,0 +1,144 @@
...
+ * Class is a bridge between EventHub KafkaConsumers and AzureEventHubsSourceTask. It verifies the configuration
+ * of kafka consumers and instantiates them, then allows AzureEventHubsSourceTask to pull for SourceRecords.
+ */
+@Slf4j
+public class EventHubsKafkaConsumerController {
+
+  private final AzureEventHubsConfig azureEventHubsConfig;
+  private final KafkaConsumerProvider kafkaConsumerProvider;
+  private final BlockingQueue<ConsumerRecords<String, String>> recordsQueue;
+  private final BlockingQueuedKafkaConsumer queuedKafkaConsumer;
```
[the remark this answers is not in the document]
GoMati-MU: This class has two purposes of now:
- It's a consumer of BlockingQueue whenever Task calls `poll()`
- It coordinates Kafka `Consumer` initialization by calling `KafkaConsumerProvider`

That's why it has both of those.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/config/AzureEventHubsConfig.java
stheppi: It's a configuration that needs to be extended. keep in mind is part of the mapping between eventhub and kafka.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureConsumerRebalancerListener.java
stheppi: Yes please

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/TopicPartitionOffsetProvider.java
stheppi: Each Task runs in its isolated class loader. So the singleton approach does not apply. I will take it offline to explain.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/BlockingQueuedKafkaConsumer.java
stheppi: How would it bubble up? :)

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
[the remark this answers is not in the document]
stheppi: A Task configuration is fixed once the Task is created. I see the logic to deal with configuration present in this class but also in KafkaConsumerProvider:

```
  KafkaConsumer<String, String> kafkaConsumer = new KafkaConsumer<>(consumerProperties);
    String topic = consumerProperties.getProperty(AzureEventHubsConfigConstants.EVENTHUB_NAME);
    return new BlockingQueuedKafkaProducer(recordBlockingQueue, kafkaConsumer, clientId, topic);
```

I think it would be better to separate and consolidate creating a Kafka consumer from the task properties, including the group and instance id. Doing so allows you to test this code in isolation, and you are not coupling the controller with the configuration either.

The KafkaConsumerProvider should give back an instance of a Consumer<Bytes, Bytes> and take care of the underlying. And in the interest of testing would it not make sense for KafkaConsumerProvider to be an interface?

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
GoMati-MU: Yeah, I addressed that

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
GoMati-MU: Unfortunately this is not true: `poll()` method can (and will!) return null record if there's nothing in the Queue, so at least on this side of the queue we have to have this check.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
GoMati-MU: just to be sure: you mean `azureEventHubsConfig.originalsWithPrefix(PREFIX, true)` method and passing it to Kafka Consumer?

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
GoMati-MU: yeah, we may want to just pass it to `KafkaConsumerProvider` depending on what you mean [here](https://github.com/lensesio/stream-reactor/pull/1033#discussion_r1508852691)

But this would delegate properties extraction which would be nice.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/TopicPartitionOffsetProvider.java
GoMati-MU: OH!!! Now I get why in [this comment](https://github.com/lensesio/stream-reactor/pull/1033#discussion_r1506151573) you say that the static won't work. I was initially under the impression that Every Task (and Connector) share the same classloader. I'll have amend it shortly, but offline explanation will definitely be appreciated!

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/BlockingQueuedKafkaProducer.java
```diff
@@ -0,0 +1,85 @@
...
+  public BlockingQueuedKafkaProducer(BlockingQueue<ConsumerRecords<String, String>> recordsQueue,
+      Consumer<String, String> consumer, String clientId, String topic) {
+    this.recordsQueue = recordsQueue;
+    this.consumer = consumer;
+    this.clientId = clientId;
+    this.topic = topic;
+  }
+
+  void start(Duration pollDuration) {
+    if (!initialized.get()) {
```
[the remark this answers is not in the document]
GoMati-MU: Since we introduced atomic boolean I think there should be no problem around that? Plus the thread is tied to class: if we want next consumer we have to create another `BlockingQueuedKafkaProducer` instance.

We can also start consuming from constructor and then not call `start()` in as part of the `poll()` anymore

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/KafkaConsumerProvider.java
GoMati-MU: Yeah, if the classloaders are different then it has to be done in the other way. I amended the code

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
GoMati-MU: clientId also needs to be taken into account

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureConsumerRebalancerListener.java
GoMati-MU: [Done](https://landoop.atlassian.net/browse/LC-128).

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/TopicPartitionOffsetProvider.java
GoMati-MU: Okay, I moved initialization to the Task, this should be way more streamlined now, thanks!

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/BlockingQueuedKafkaConsumer.java
GoMati-MU: From what I can see `consumer.poll()` method throws `RuntimeException`s so those will not get catched and will result in stacktrace.

Alternative would be to wrap all of them into something like `ConsumptionPollingException` and log it, so I can also go down that path

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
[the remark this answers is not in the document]
GoMati-MU: Agreed on removing code that dealt with properties from here and delegated to `ProducerProvider`.
This also means we don't have to keep some of those fields and class if more in-line with _Single Responsibility Principle_.

About interface: I added it as it will pay off when we abstract from String (and if we ever abstract from Kafka to let's say Azure SDK)

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
GoMati-MU: delegated this to `ProducerProvider` like advised, thanks!

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/BlockingQueuedKafkaProducer.java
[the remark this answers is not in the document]
GoMati-MU: `start()` is now part of the constructor.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/EventHubsKafkaConsumerController.java
GoMati-MU: This was delegated to `BlockingQueueProducerProvider`

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/reading/BlobCheckpointStoreSample.java
```diff
@@ -0,0 +1,94 @@
+//package io.lenses.java.streamreactor.connect.azure.eventhubs.reading;
```
stheppi, line 1: why not delete this code?

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/reading/EventProcessorBlobCheckpointStoreSample.java
```diff
@@ -0,0 +1,71 @@
+//package io.lenses.java.streamreactor.connect.azure.eventhubs.reading;
```
stheppi, line 1: Why not remove this since it's not used?

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureEventHubsSourceTask.java
```diff
@@ -28,29 +41,36 @@ public String version() {
...
+    initialize(eventHubsKafkaConsumerController);
+  }
+
+  /**
+   * Initializes the Task. This method shouldn't be called if start() was already called with
+   * {@link EventHubsKafkaConsumerController} instance.
+   *
+   * @param eventHubsKafkaConsumerController {@link EventHubsKafkaConsumerController} for this task
+   */
+  public void initialize(EventHubsKafkaConsumerController eventHubsKafkaConsumerController) {
```
stheppi, line 58: Is this meant to help with testing?

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureEventHubsSourceTask.java
```diff
@@ -28,29 +41,36 @@ public String version() {
...
-      configProps = props;
-    } else {
-      assert context != null;
-      configProps = context.configs();
-    }
+    eventHubsKafkaConsumerController = new EventHubsKafkaConsumerController(
+        new AzureEventHubsConfig(props), new BlockingQueueProducerProvider(),
+        new ArrayBlockingQueue<>(RECORDS_QUEUE_DEFAULT_SIZE));
+    ofNullable(this.context).flatMap(context -> ofNullable(context.offsetStorageReader()))
+        .ifPresent(TopicPartitionOffsetProvider::initialize);
```
stheppi, line 48: Why do we need to have the call to this initialise with all its synchronization when the call should happen once? Why not create a new TopicPartitionOffsetProvider(OffsetStorageReader) since this class is a soft wrapper to offer a few helper methods?  Why is the advantage of even having the wrapper over static functions.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureEventHubsSourceTask.java
GoMati-MU: Yes, this is to ommit `context()` calls that are part of `start()` method in tests.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/reading/EventProcessorBlobCheckpointStoreSample.java
GoMati-MU: my bad, thanks for checking! deleted now

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureConsumerRebalancerListener.java
```diff
@@ -0,0 +1,59 @@
...
+  Consumer<?, ?> kafkaConsumer;
+
+  public AzureConsumerRebalancerListener(
+      Optional<TopicPartitionOffsetProvider> topicPartitionOffsetProvider,
+      Consumer<?, ?> kafkaConsumer) {
+    this.topicPartitionOffsetProvider = topicPartitionOffsetProvider;
+    this.kafkaConsumer = kafkaConsumer;
+  }
+
+  public AzureConsumerRebalancerListener(Consumer<?, ?> kafkaConsumer) {
```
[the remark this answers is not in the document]
GoMati-MU: Simple - for testing. It's either that or I will have to mock static calls (which is a way straight down to hell ;) )

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureEventHubsSourceTask.java
GoMati-MU: singletonb drawio

Please consider above diagram. The only class that uses singleton `TopicPartitionOffsetProvider` (indicated red) is `AzureConsumerRebalancerListener` (indicated blue) but in order for it to be initialize singleton we have to pass it `TopicPartitionOffsetProvider` that only `AzureEventHubsSourceTask` has (purple color). So if we resign from current singleton approach we will have to do one of the following:
- pass `TopicPartitionOffsetProvider` through all of the black classes even though they don't really use it
- still initialize `TopicPartitionOffsetProvider` in Task but pass it down just like above
- create `AzureConsumerRebalancerListener` in Task (even though it doesn't care) and pass it all the way to Kafka Consumer

IMO it's way better to do initialization like it's done right now: initialization will only happen once and classes that have nothing to do with `TopicPartitionOffsetProvider` won't have to deal with it at all.

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/reading/BlobCheckpointStoreSample.java
GoMati-MU: my bad, thanks for checking! deleted now

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureEventHubsSourceTask.java
stheppi: >pass TopicPartitionOffsetProvider through all of the black classes even though they don't really use it

Isn't the class hierarchy structure causing this? I find the singleton thread-safety initialisation an overhead.
The ConsumerController can get the BlockingQueueProducer injected, and it's not its responsibility to create it. And that can happen in the Task::start method. The connector configuration has all the information to do it. 
The ProducerProvider can just be:

```java
public interface ProducerProvider {

  BlockingQueueProducer createProducer(BlockingQueue<ConsumerRecords<String, String>> recordBlockingQueue);
}
```

Then the Controller does not have to know about the azureconfig and the need for the Task::context.storageOffseet being passed down through so many layers won't be an "issue". 



Also keep in mind that BlockingQueuedKafkaProducer receives the `Consumer<string, string>` created on a thread different from the one that calls read. See the docs [here](https://kafka.apache.org/36/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#multithreaded), you see the subscribe happens on the thread consuming?

GoMati-MU: After our conversation I changed it to resign from Singleton approach

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureConsumerRebalancerListener.java
```diff
@@ -0,0 +1,48 @@
...
+  @Override
+  public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
+    partitions.forEach(partition -> {
+      AzureTopicPartitionKey partitionKey = new AzureTopicPartitionKey(
+          partition.topic(), partition.partition());
+      Optional<AzureOffsetMarker> partitionOffset = topicPartitionOffsetProvider.getOffset(partitionKey);
+      partitionOffset.ifPresentOrElse(
+          offset -> kafkaConsumer.seek(partition, offset.getOffsetValue()),
+          () -> kafkaConsumer.seekToBeginning(Collections.singletonList(partition)));
+      log.info("Subscribed to topic: {}", partition.topic());
```
stheppi, line 44: Minor one.
If the topic has 8 partitions this will be logged 8 times :)

stheppi (approved):

java-connectors/kafka-connect-azure-eventhubs/src/main/java/io/lenses/java/streamreactor/connect/azure/eventhubs/source/AzureConsumerRebalancerListener.java
GoMati-MU: amended, thanks! :)

PR #1038, opened 2024-03-06
Partition Search

davidsloan:
This introduces most of the functionality for the GCP Storage Source.

* Removes `DirectoryFindCompletionConfig` and `DirectoryFindResults` case classes as these were simple wrappers for primitive values and adding indirection to the code.
* Some functionality is pushed up from `S3SourceTask` into `CloudSourceTask` to be shared by all cloud connectors, with some abstract template methods defined in `CloudSourceTask` to ensure the functionality is implemented by all connectors.
* `S3PartitionSearcher` is promoted to cloud-commons module and renamed to `CloudPartitionSearcher`.
* `DirectoryLister` becomes a trait (interface) sitting in cloud-commons to be implemented by each source subproject.  Implementations of this trait are `GCPStorageDirectoryLister` and `AWSS3DirectoryLister`.
* `GCPStorageSourceTask` and `GCPStorageSourceConnector` are introduced but not yet ready to use until some methods in the `GCPStorageStorageInterface` are implemented.

### DIFF
--- a/kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/cloud/common/utils/AsyncIOAssertions.scala
+++ b/kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/cloud/common/utils/AsyncIOAssertions.scala
@@ -1,0 +1,17 @@
+package io.lenses.streamreactor.connect.cloud.common.utils
+
+import cats.effect.IO
+
+/**
+  * This is a workaround for being unable to use AsyncIOSpec in some tests due to their superclasses using some other non-IO spec.
+  */
+trait AsyncIOAssertions {
+
+  implicit class IOAssertionOps[T](io: IO[T]) {
+    def asserting(predicate: T => Unit): IO[T] = io.map { value =>
+      predicate(value)
+      value
+    }
+  }
+
+}
--- a/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3DirectoryLister.scala
+++ b/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3DirectoryLister.scala
@@ -44,12 +44,12 @@ class AwsS3DirectoryLister(connectorTaskId: ConnectorTaskId, s3Client: S3Client)
     wildcardExcludes: Set[String],
   ): IO[Set[String]] =
     for {
-      iterator   <- listObjects(bucketAndPrefix)
+      iterator   <- listObjects(filesLimit, bucketAndPrefix)
       prefixInfo <- extractPrefixesFromResponse(iterator, exclude, wildcardExcludes, recurseLevels)
       flattened  <- flattenPrefixes(bucketAndPrefix, filesLimit, prefixInfo, recurseLevels, exclude, wildcardExcludes)
     } yield flattened
 
-  private def listObjects(bucketAndPrefix: CloudLocation): IO[Iterator[ListObjectsV2Response]] = {
+  private def listObjects(filesLimit: Int, bucketAndPrefix: CloudLocation): IO[Iterator[ListObjectsV2Response]] = {
 
     def createListObjectsRequest(
       bucketAndPrefix: CloudLocation,
@@ -57,7 +57,7 @@ class AwsS3DirectoryLister(connectorTaskId: ConnectorTaskId, s3Client: S3Client)
 
       val builder = ListObjectsV2Request
         .builder()
-        .maxKeys(1000)
+        .maxKeys(filesLimit)
         .bucket(bucketAndPrefix.bucket)
         .delimiter("/")
       bucketAndPrefix.prefix.foreach(builder.prefix)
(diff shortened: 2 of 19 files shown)


kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/cloud/common/utils/AsyncIOAssertions.scala
```diff
@@ -0,0 +1,17 @@
+package io.lenses.streamreactor.connect.cloud.common.utils
+
+import cats.effect.IO
+
+/**
+  * This is a workaround for being unable to use AsyncIOSpec in some tests due to their superclasses using some other non-IO spec.
```
stheppi, line 6: should we separate the IO tests?

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3DirectoryLister.scala
```diff
@@ -20,94 +20,79 @@ import cats.implicits._
...
+
+  private def listObjects(bucketAndPrefix: CloudLocation): IO[Iterator[ListObjectsV2Response]] = {
+
+    def createListObjectsRequest(
+      bucketAndPrefix: CloudLocation,
+    ): ListObjectsV2Request = {
+
+      val builder = ListObjectsV2Request
+        .builder()
+        .maxKeys(1000)
```
stheppi, line 59: I thought the connector BATCH was controlling the number of keys. If so we should not hardcode the value

kafka-connect-gcp-storage/src/main/scala/io/lenses/streamreactor/connect/gcp/storage/storage/GCPStorageDirectoryLister.scala
```diff
@@ -0,0 +1,80 @@
...
+    recurseLevels:    Int,
+    exclude:          Set[String],
+    wildcardExcludes: Set[String],
+  ): IO[Set[String]] = {
+
+    def listSubdirs(prefix: String, recurseLevels: Int): Iterable[String] = {
+
+      val blobListOptions = BlobListOption.dedupe(
+        BlobListOption.delimiter("/"),
+        BlobListOption.pageSize(1000),
```
stheppi, line 46: Similar comment around the page size and KCQL BATCH setting.

kafka-connect-gcp-storage/src/main/scala/io/lenses/streamreactor/connect/gcp/storage/storage/GCPStorageDirectoryLister.scala
```diff
@@ -0,0 +1,80 @@
...
+    * @param wildcardExcludes allows ignoring paths containing certain strings.  Mainly it is used to prevent us from reading anything inside the .indexes key prefix, as these should be ignored by the source.
+    */
+  override def findDirectories(
+    bucketAndPrefix:  CloudLocation,
+    recurseLevels:    Int,
+    exclude:          Set[String],
+    wildcardExcludes: Set[String],
+  ): IO[Set[String]] = {
+
+    def listSubdirs(prefix: String, recurseLevels: Int): Iterable[String] = {
```
stheppi, line 42: I think this method can be made tail-recursive

```scala
def listSubdirs(prefix: String, recurseLevels: Int): Iterable[String] = {

  @tailrec
  def loop(prefixes: List[String], acc: List[String]): Iterable[String] = {
    if (prefixes.isEmpty || recurseLevels <= 0) {
      acc
    } else {
      val foundResults = prefixes.flatMap { p =>
        val blobListOptions = BlobListOption.dedupe(
          BlobListOption.delimiter("/"),
          BlobListOption.pageSize(1000),
          BlobListOption.prefix(p),
          BlobListOption.currentDirectory(),
        )

        storage
          .get(bucketAndPrefix.bucket)
          .list(blobListOptions: _*)
          .iterateAll()
          .asScala
          .filter(_.isDirectory)
          .map(_.getName)
          .toList
          .filter { prefix =>
            connectorTaskId.ownsDir(prefix) && !exclude.contains(prefix) && !wildcardExcludes.exists(we =>
              prefix.contains(we),
            )
          }
      }

      val nextPrefixes = foundResults.filter(_ => recurseLevels > 1)

      loop(nextPrefixes, acc ++ foundResults)
    }
  }

  loop(List(prefix), List())
}
```

i haven't validated it, write on paper kind of approach.

kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/cloud/common/utils/AsyncIOAssertions.scala
davidsloan: I am not sure the best approach.  Ideally we could revisit `CloudPlatformEmulatorSuite` and upgrade it to the IO world, but it has a lot of subclasses that we'd also have to touch.  So I avoided it and did this 😁

kafka-connect-gcp-storage/src/main/scala/io/lenses/streamreactor/connect/gcp/storage/storage/GCPStorageDirectoryLister.scala
davidsloan: Love it, and wanted to use it, but couldn't get it working 😢

stheppi (approved):

PR #1048, opened 2024-03-13
Chore/dep tidy

davidsloan:
* Some classes in `kafka-connect-common` are only used in a single module, so these move to their module.
* Removes some dependencies unused in the project (eg guava, snakeyaml)
* Reduces some of the noise in KcqlSettings by extracting what's common to the sql modules only into KcqlWithFieldSettings
* KCQL has only 1 entry point in the scala code.
* Ensure that KCQL.query is populated as this is used for MQTT.


GoMati-MU (commented):
small comment, beside that I think the solution nicely streamlines the logic

GoMati-MU (approved):
Approved, seems like a solid solution

PR #1049, opened 2024-03-13
Upgrade hadoop to avoid CVE-2022-25168

davidsloan:


andrewstevenson (approved):

project/Dependencies.scala
```diff
@@ -42,22 +42,22 @@ object Dependencies {
 
   object Versions {
     // libraries versions
-    val scalatestVersion               = "3.2.15" // Higher versions result in org.scala-lang.modules:scala-xml conflicts
```
GoMati-MU, line 45: is this still the case?

davidsloan: No, so I took the comment out 😁

PR #1050, opened 2024-03-14
Fix discrepancy between code and docs for recurse levels/millis/mode

davidsloan:
The following properties are described in the docs:

```
connect.s3.source.partition.search.continuous    
connect.s3.source.partition.search.interval      
connect.s3.source.partition.search.recurse.levels
```

And these follow the convention of the other properties.  However in implementation the properties are:

```
connect.s3.partition.search.continuous    
connect.s3.partition.search.interval      
connect.s3.partition.search.recurse.levels
```

This PR corrects the code to reflect the properties in the docs.  The `DeprecationConfigDefProcessor` will also cause an descriptive error on startup if the old property continues to be used, enabling the user to update their configuration to the correct one.

The error in the logs will look like this:

```
Change `connect.s3.partition.search.continuous` to `connect.s3.source.partition.search.continuous`
```


GoMati-MU (approved):
Seems solid, approved from me

PR #1056, opened 2024-03-20
fix for KCQL compilates to appear in the release (ambiguous BASH behaviour)

GoMati-MU:
Fix for KCQL compilates that didn't get picked up for release in the first place because of github using `bash` (where `*` and `**` are treated the same by default, not like e.g. `zsh`)


davidsloan (approved):
LGTM

PR #1103, opened 2024-04-02
remove 'java' part from package name

GoMati-MU:
#### Description

This PR is to get rid of `java` part in java connectors code packages. When checking we have to pay attention if everything has been moved (as far as I can see this is the case but let me know if it's not).


java-connectors/settings.gradle
```diff
@@ -2,5 +2,6 @@ rootProject.name = 'java-reactor'
 include 'kafka-connect-common',
 'kafka-connect-azure-eventhubs',
 'kafka-connect-cloud-common',
+'kafka-connect-mongodb',
```
davidsloan, line 5: ???

davidsloan (approved):
One small issue I noticed, why is mongodb referenced in this file.  Also might as well check the Codacy issue that's appeared.

java-connectors/settings.gradle
GoMati-MU: good spot! Removed, there's no need for that :)

PR #1105, opened 2024-04-04
adjust scripts to allow gradle scripts for easy JARs release

GoMati-MU:
#### Description
Small adjustment to build scripts to output them nicely in following way:

image


GoMati-MU:
Added to: https://github.com/lensesio/stream-reactor/pull/1108

PR #1109, opened 2024-04-08
Using connector name as consumer group

GoMati-MU:
#### Description

This short PR is to use Connector's Name (picked from config `name`) as consumer group for Kafka Consumer instead of relying on user to provide it.


stheppi (changes requested):
I think there is more to this PR to be done.

The connector code appends the Kafka consumer settings with a prefix. That code needs to be updated to drop the `group.id` setting such that it's not even a setting that the user can set.

GoMati-MU:
> > I think there is more to this PR to be done.
> > The connector code appends the Kafka consumer settings with a prefix. That code needs to be updated to drop the `group.id` setting such that it's not even a setting that the user can set.
> 
> Unfortunately I cannot reply directly under comment so I reply here.
> 
> `group.id` was not added to Connector's ConfigDef as it was one of the passthrough configs (we were basically stripping prefix from it) so we don't have to delete it from there. Basically what needs to be done is to add another parameter (`group.id`) to documentation to indicate to user that it's being specified by the connector and that it's being ignored and not being passed through.
> 
> Happy to discuss if needed

After discussion we've decided to remove them also from ConfigDef.

cmcmteixeira (approved):

PR #1137, opened 2024-04-12
LC-187 Avoid adapting the partition fields value

stheppi:
At the moment an SMT can control the format of the key. For example it can set it to: yyyy-MM-dd/HH. But the connector replaces the `/` and thus breaks the key.

The change removes the code to replace the value and adds a test for it.

### DIFF
--- a/kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/naming/CloudKeyNamer.scala
+++ b/kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/naming/CloudKeyNamer.scala
@@ -199,10 +199,7 @@ class CloudKeyNamer(
 
   private def getFieldStringValue(struct: SinkData, partitionName: Option[PartitionNamePath]) =
     adaptErrorResponse(SinkDataExtractor.extractPathFromSinkData(struct)(partitionName)).fold(Option.empty[String])(
-      fieldVal =>
-        Option(fieldVal
-          .replace("/", "-")
-          .replace("\\", "-")),
+      fieldVal => Option(fieldVal),
     )
 
   private def getPartitionValueFromSinkData(sinkData: SinkData, partitionName: PartitionNamePath): String =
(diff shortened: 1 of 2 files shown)


kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/naming/CloudKeyNamer.scala
```diff
@@ -199,10 +199,7 @@ class CloudKeyNamer(
 
   private def getFieldStringValue(struct: SinkData, partitionName: Option[PartitionNamePath]) =
     adaptErrorResponse(SinkDataExtractor.extractPathFromSinkData(struct)(partitionName)).fold(Option.empty[String])(
-      fieldVal =>
-        Option(fieldVal
-          .replace("/", "-")
-          .replace("\\", "-")),
+      fieldVal => Option(fieldVal),
```
davidsloan, line 202: Actually you can simplify this further!

adaptErrorResponse(SinkDataExtractor.extractPathFromSinkData(struct)(partitionName))

You don't need the fold or anything after

kafka-connect-cloud-common/src/test/scala/io/lenses/streamreactor/connect/cloud/common/sink/naming/CloudKeyNamerTest.scala
```diff
@@ -69,6 +70,30 @@ class CloudKeyNamerTest extends AnyFunSuite with Matchers with OptionValues with
...
+        topicPartition.offset,
+      ),
+      topicPartition.toTopicPartition,
+    )
+
+    either match {
+      case Left(_) => fail("Should not have failed")
+      case Right(value) =>
+        value shouldBe Map(HeaderPartitionField(PartitionNamePath("h")) -> "val1/val2")
+    }
```
davidsloan, line 94: I would probably just use `EitherValues` trait which is already being used

`either.value should be Map(HeaderPartitionField(PartitionNamePath("h")) -> "val1/val2")

Then no need for the conditional match 😁

davidsloan (approved):

PR #1145, opened 2024-04-15
java connectors GitHub workflows

GoMati-MU:
#### Description

Those changes to `Gradle` build and new scripts to GitHub Actions should allow `sbt` and `Gradle` builds to work simultaneously in order to allow for both java and scala connectors to be released alongside. This should also provide the base for further modules to be build by java and used by Scala.

#### Testing
Below testing was done:

- [x] Tests of Gradle build script changes locally
- [x] Github actions tests on my fork ( https://github.com/GoMati-MU/stream-reactor )

### DIFF
--- a/.github/workflows/java-build.yml
+++ b/.github/workflows/java-build.yml
@@ -59,7 +59,7 @@ jobs:
       - name: Test with Gradle
         run: cd 'java-connectors' && ./gradlew ${{ matrix.module }}:test
 
-  build-and-upload-artifact:
+  build-and-cache:
     needs:
       - test
       - initiate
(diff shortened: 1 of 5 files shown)


.github/workflows/java-build.yml
```diff
@@ -0,0 +1,99 @@
...
+    strategy:
+      matrix:
+        module: ${{fromJSON(needs.initiate.outputs.matrix)}}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK 21
+        uses: actions/setup-java@v4
+        with:
+          java-version: 21
```
davidsloan, line 75: https://docs.confluent.io/platform/current/installation/versions-interoperability.html

We can't build the release with jdk 21, 17 is the maximum for wide support unfortunately.

GoMati-MU: Changed, thank you! 
(I was originally thinking setting bytecode to 17 in Gradle will be enough, but let's stay on the safe side)

GoMati-MU:
moved to https://github.com/lensesio/stream-reactor/pull/1147

PR #1155, opened 2024-04-16
LC-191 Workaround  Kafka Connect SMTs returning POJO

stheppi:
Some users have implemented SMTs that deviate from recommended best practices, returning Plain Old Java Objects (POJOs) instead of constructing Connect structs. Consequently, such SMTs encounter compatibility issues with the JsonConverter, which is designed to handle Connect data structures.

To address this, the proposed changes implement a mechanism to detect the presence of POJOs within the pipeline. Upon detection, the system leverages the Jackson Json writer to marshal the payload into JSON format, ensuring compatibility with the JsonConverter.

This code would only work for Json storage format and not for Avro/Parquet. We don't plan workarounds for those scenarios

### DIFF
--- a/kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/conversion/ToJsonDataConverter.scala
+++ b/kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/conversion/ToJsonDataConverter.scala
@@ -37,20 +37,13 @@ object ToJsonDataConverter {
   def convertMessageValueToByteArray(converter: JsonConverter, topic: Topic, data: SinkData): Array[Byte] =
     data match {
       case data: PrimitiveSinkData => converter.fromConnectData(topic.value, data.schema().orNull, data.safeValue)
-      case StructSinkData(structVal)    => converter.fromConnectData(topic.value, data.schema().orNull, structVal)
-      case MapSinkData(map, schema)     => converter.fromConnectData(topic.value, schema.orNull, map)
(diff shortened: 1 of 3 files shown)


kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/conversion/ToJsonDataConverter.scala
```diff
@@ -23,21 +25,35 @@ import io.lenses.streamreactor.connect.cloud.common.formats.writer.PrimitiveSink
...
+        val isPojo = array.size() > 0 && array.asScala.exists {
+          case _: Struct => false
+          case _ => true
+        }
+        if (isPojo) {
+          val json = jacksonJson.writeValueAsString(array)
+          json.getBytes()
+        } else {
+          converter.fromConnectData(topic.value, schema.orNull, array)
+        }
```
davidsloan, line 53: What do you think of refactoring this a little to break it up, introducing a private method that could be documented with why this is a hack and should be avoided:

```suggestion
      case ArraySinkData(array, _) if isPojo(array) =>
          val json = jacksonJson.writeValueAsString(array)
          json.getBytes()
      case ArraySinkData(array, schema) =>
          converter.fromConnectData(topic.value, schema.orNull, array)
```


```suggestion
  /**
   * This is a workaround to help some of the customers who use Kafka Connect SMT ignoring the best practices
   */
  private def isPojo(array: util.List[_]) = {
    array.size() > 0 && array.asScala.exists {
      case _: Struct => false
      case _ => true
    }
  }
```

davidsloan (approved):

PR #1175, opened 2024-04-24
Enabling java module in scala via sbt

davidsloan:


GoMati-MU:
#### Description

This is to address technical debt introduced by Java Connectors and separate build. Now we rewire some older common utils to their Java counterparts.
Authors: @davidsloan (sbt build part), @GoMati-MU (rewiring, some small changes)

#### Testing
Full regression as there was no logic change here.

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
```diff
@@ -53,12 +53,16 @@ public class JarManifest {
   public JarManifest(URL location) {
     Manifest manifest;
 
-    try (JarFile jarFile = new JarFile(new File(location.toURI()))) {
-      manifest = jarFile.getManifest();
+    try {
+      File file = new File(location.toURI());
+      if (file.isFile()) {
+        JarFile jarFile = new JarFile(file);
```
davidsloan, line 59: resource `jarFile` is never closed.

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
```diff
@@ -69,11 +73,13 @@ public JarManifest(JarFile jarFile) {
     Manifest manifest;
     try {
       Optional<JarFile> jarFileOptional = of(jarFile);
-      manifest = jarFileOptional.get().getManifest();
+      if (jarFileOptional.isPresent()) {
```
davidsloan, line 76: this check is redundant, as the Option will be present as we are using `of()` instead of `ofNullable()`

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
```diff
@@ -69,11 +73,13 @@ public JarManifest(JarFile jarFile) {
...
+        manifest = jarFileOptional.get().getManifest();
+        extractMainAttributes(manifest.getMainAttributes());
+      }
     } catch (NullPointerException | IOException e) {
```
davidsloan, line 80: We shouldn't really need to be catching a NPE if we are already using an Optional?

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
```diff
@@ -69,11 +73,13 @@ public JarManifest(JarFile jarFile) {
...
       throw new ConnectorStartupException(e);
     }
-    extractMainAttributes(manifest.getMainAttributes());
   }
 
   private void extractMainAttributes(Attributes mainAttributes) {
```
davidsloan, line 85: Food for thought : 🍕 

When a function returns void, it typically indicates that the function performs side effects (such as modifying mutable state, printing to the console, or writing to a file) without producing any meaningful result. Such functions are not referentially transparent because their effects are observable and can change the behaviour of subsequent function calls.

instead of this returning void, it would be good for referential transparancy if this constructed the map immutably and returned the new map, to be set into the class field (which could also be an Optional<>, empty when uninitialised.

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
```diff
@@ -69,11 +73,13 @@ public JarManifest(JarFile jarFile) {
     Manifest manifest;
     try {
       Optional<JarFile> jarFileOptional = of(jarFile);
-      manifest = jarFileOptional.get().getManifest();
+      if (jarFileOptional.isPresent()) {
+        manifest = jarFileOptional.get().getManifest();
```
davidsloan, line 77: It would be nice to avoid the `.get` on the Optional here.

stheppi: +1 on working with Optionl

```java
of(jarFile).flatMap(j-> of(j.getManifest)).flatMap(m->of(m.getMainAttributes).map(extractMainAttributes).orNull
```

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
GoMati-MU: it is now :) thanks!

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
GoMati-MU: yup, I changed that

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
GoMati-MU: things were a little bit more complicated here as it's `getManifest()` that could give us null. Still, it wasn't neat so I changed that.

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
GoMati-MU: Good one! I used this idea for `extractMainAttributes` and `buildManifestString` function as well!

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
GoMati-MU: Unfortunately it's not that easy as `getManifest()` can give us `IOException` which we probably wouldn't like to try to catch inside of functional stream. Still, I tried to make it as functional as possible, which I hope will be a nice consensus (it'll become a monad but there will be no try-catch inside) between our ideas.

kafka-connect-elastic7/src/main/scala/io/lenses/streamreactor/connect/elastic7/ElasticSinkConnector.scala
```diff
@@ -32,7 +32,7 @@ import scala.jdk.CollectionConverters.SeqHasAsJava
 class ElasticSinkConnector extends SinkConnector with StrictLogging {
   private var configProps: Option[util.Map[String, String]] = None
   private val configDef = ElasticConfig.config
-  private val manifest  = JarManifest(getClass.getProtectionDomain.getCodeSource.getLocation)
+  private val manifest  = new JarManifest(getClass.getProtectionDomain.getCodeSource.getLocation)
```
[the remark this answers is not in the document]
GoMati-MU: Neat idea, let me try that!

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
davidsloan: Sorry, no, there's still a potential issue with the jarFile.close() call. If an exception occurs before jarFile is closed, the file won't be closed properly, potentially leading to resource leaks. To ensure proper resource management, you should use a try-with-resources statement

```
try (JarFile jarFile = new JarFile(file)) {
                jarAttributes = Optional.ofNullable(jarFile.getManifest())
                        .flatMap(mf -> Optional.of(mf.getMainAttributes()))
                        .map(this::extractMainAttributes)
                        .orElseThrow(() -> new ConnectorStartupException("Failed to extract main attributes from Jar file."));
            }
```

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
```diff
@@ -44,51 +45,50 @@ public class JarManifest {
...
-        ofNullable(mainAttributes.getValue(REACTOR_DOCS.getAttributeName())).orElse(UNKNOWN));
+
+  private Map<String, String> extractMainAttributes(Attributes mainAttributes) {
+    final Map<String, String> attributesFromJar = new HashMap<>();
+    Arrays.stream(ManifestAttributes.values()).forEach(manifestAttribute ->
+        attributesFromJar.put(manifestAttribute.getAttributeName(),
+            ofNullable(mainAttributes.getValue(manifestAttribute.getAttributeName())).orElse(UNKNOWN))
+    );
+    return attributesFromJar;
   }
```
davidsloan, line 92: 🚀 

How do you feel about the below tweak to avoid the mutable map state?

```suggestion
  private Map<String, String> extractMainAttributes(Attributes mainAttributes) {
    final Map<String, String> attributesFromJar = new HashMap<>();
    Arrays.stream(ManifestAttributes.values()).forEach(manifestAttribute ->
        attributesFromJar.put(manifestAttribute.getAttributeName(),
            ofNullable(mainAttributes.getValue(manifestAttribute.getAttributeName())).orElse(UNKNOWN))
    );
    return attributesFromJar;
  }
```

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
```diff
@@ -103,18 +103,12 @@ public String getVersion() {
...
-        .append(jarAttributes.get(KAFKA_VER.getAttributeName())).append(NEW_LINE);
-    manifestBuilder.append(GIT_REPO.attributeName).append(SEMICOLON)
-        .append(jarAttributes.get(GIT_REPO.getAttributeName())).append(NEW_LINE);
-    manifestBuilder.append(GIT_HASH.attributeName).append(SEMICOLON)
-        .append(jarAttributes.get(GIT_HASH.getAttributeName())).append(NEW_LINE);
-    manifestBuilder.append(GIT_TAG.attributeName).append(SEMICOLON)
-        .append(jarAttributes.get(GIT_TAG.getAttributeName())).append(NEW_LINE);
-    manifestBuilder.append(REACTOR_DOCS.attributeName).append(SEMICOLON)
-        .append(jarAttributes.get(REACTOR_DOCS.getAttributeName())).append(NEW_LINE);
+    List<ManifestAttributes> attributesInStringOrder =
```
davidsloan, line 106: 🚀

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
GoMati-MU: I'm indifferent to for it to be immutable to be honest: I don't see the requirement, nor I have anything against in this scenario :)
I've changed that using your suggestion

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/JarManifest.java
GoMati-MU: Please, don't be sorry, since it's a totally valid argument, it still can be left open in scenario you've mentioned.

Or could be, but I patched it, now it _should_ be alright ;)

stheppi (approved):

PR #1178, opened 2024-04-25
Fix the release build

stheppi:
previous variable was only set for a PR, however the release uses a tag trigger. This time it hardcodes lensesio


davidsloan (approved):
Approving for the release.

this will break the scala steward builds, so let’s set it back after the release.

PR #1184, opened 2024-04-26
Enhancements for Source Line-Start-End Functionality (LC-196)

stheppi:
This PR addresses a specific challenge encountered by users of the S3 source functionality. In cases where an external producer terminates a file without the end message mark. This ends up losing data.

To mitigate this issue, the PR introduces a new property entry for KCQL to signal the unterminated message to be considered.


davidsloan (approved):

kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/config/kcqlprops/PropsKeyEnum.scala
```diff
@@ -37,6 +37,8 @@ object PropsKeyEnum extends Enum[PropsKeyEntry] {
   case object ReadStartLine extends PropsKeyEntry("read.text.start.line")
   case object ReadEndLine   extends PropsKeyEntry("read.text.end.line")
 
+  case object ReadLastEndLineMissing extends PropsKeyEntry("read.text.last.end.line.missing")
```
GoMati-MU, line 40: MINOR: Could be named `ReadDespiteLastEndLineMissing` for more clarity (IMO), but it's not a blocker IMO

stheppi: I don't see how it's clear. 

* Read is the prefix for text related source functionality
* LastEndLineMissing is a flag (true/false)

PR #1187, opened 2024-04-29
S3/GCS/Azure Source: Enhanced Data Reload Strategy for Specific Timeframes

stheppi:
### Overview
This pull request addresses the challenge of efficiently reloading data from a specific point in time (e.g., last week, yesterday). Currently, the naive approach involves loading data from the beginning and filtering out records, which proves impractical and costly, especially with large data volumes.

#### Proposed Solution
To enhance data reloads for specific timeframes, this change introduces a more refined strategy. Instead of loading all stored data and filtering records, the solution leverages timestamp metadata associated with each data file. By utilizing this metadata, the source can selectively load relevant data files, minimizing unnecessary data transfer and processing overhead.

This PR prepares the key (filename) values to carry the earliest and latest offset for the records within each entry. This allows the source to discard files that should not be processed, making the initial seek time directly proportional to the listing of keys and associated filters for each topic-partition. Additionally, records returned from the file undergo further filtering, ensuring only relevant data within the specified timeframe is processed.

## Changes

This change exclusively affects the envelope storage mode. In non-envelope storage, there is no inherent storage mechanism for the record timestamp. However, to maintain consistency across storage modes, we've introduced the TopicPartitionOffsetFileNamerV1.

It's important to note that the lexicographic order remains unaffected when storing both envelope and non-envelope data. This is because the padded-offset continues to determine the order in both scenarios.


kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/source/TempFileHelper.scala
```diff
@@ -0,0 +1,23 @@
+package io.lenses.streamreactor.connect.aws.s3.source
+
+import java.io.File
+import java.util.UUID
+
+trait TempFileHelper {
```
davidsloan, line 6: Would be nice if this used `java.nio.file.Files.createTempDirectory` or  `createTempFile` and then call `deleteOnExit()`
as this will automatically do the cleanup when the jvm is terminated (in case, for whatever reason, the cleanup is not successful).

I'm often noticing files accumulated generated during tests.

kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/writer/WriteState.scala
```diff
@@ -32,48 +32,57 @@ case class NoWriter(commitState: CommitState) extends WriteState(commitState) wi
...
+  commitState:             CommitState,
+  formatWriter:            FormatWriter,
+  file:                    File,
+  uncommittedOffset:       Offset,
+  earliestRecordTimestamp: Long,
+  latestRecordTimestamp:   Long,
 ) extends WriteState(commitState)
     with LazyLogging {
 
   //TODO: it's not clear why we are only keeping track of one schema (VALUE) and not key/and headers
```
davidsloan, line 53: Is this TODO still relevant?

davidsloan (approved):

kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/writer/WriteState.scala
stheppi: i don't think so.

kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/source/TempFileHelper.scala
stheppi: createTempDirectory does not exist, and my understanding is that this creates files and not directories. 
The code was guarding for deletion, but I added the .deleteOnExit .

PR #1193, opened 2024-05-03
Java Automatic Formatting (Config Only)

davidsloan:
* Adds autoformatting via [Spotless](https://github.com/diffplug/spotless) and tweaks the buld to enforce formatting.  This also takes care of the license headers, too, in place of the other plugin.
    * `sbt formatAll` command has been updated so that it will run the Spotless Gradle format.
    * to just reformat java code you can run `gradle spotlessApply`.
    * TODO: Align team IDE formats/preferred styles.

This PR only has the config changes to enable formatting.  No autoformatting has been applied at this point.  This allows us to define the style we want before formatting.

The formatting style can be tweaked using [the configuration in the gradle file](https://github.com/lensesio/stream-reactor/pull/1193/files#diff-0792be191ed6b0e384ee253ea969a436ff2810da6050c1dc85b5e40a915b72e4R144-R148).

Currently it is Google's style with minor tweaks.

It is not expected that this PR will be able to build until the formatting changes have been applied, as it adds the format checking to the build.

### DIFF
--- a/.github/workflows/java-build.yml
+++ b/.github/workflows/java-build.yml
@@ -56,7 +56,10 @@ jobs:
         with:
           gradle-version: 8.6
 
-      - name: Check License Headers and Test with Gradle
+      - name: Formatting and Headers Check
+        run: cd 'java-connectors' && ./gradlew ${{ matrix.module }}:spotlessCheck
+
+      - name: Test with Gradle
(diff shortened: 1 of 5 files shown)


davidsloan:
Superceded by #1210

PR #1219, opened 2024-05-12
LC-200 Skip Glacier stored objects

stheppi:
The path of listing keys was missed. This meant that for scenarios where the S3 is read with a LastModified order, the change was not working.

It also removes the logging because it is extensive with every poll of the keys.

### DIFF
--- a/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3StorageFilter.scala
+++ b/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3StorageFilter.scala
@@ -25,8 +25,10 @@ import software.amazon.awssdk.services.s3.model.S3Object
   */
 object AwsS3StorageFilter {
 
-  def filterOut(o: S3Object) =
-    o.storageClass() == ObjectStorageClass.GLACIER ||
-      o.storageClass() == ObjectStorageClass.DEEP_ARCHIVE ||
-      o.storageClass() == ObjectStorageClass.GLACIER_IR
+  private val archiveStorageClasses = Set(
+    ObjectStorageClass.GLACIER,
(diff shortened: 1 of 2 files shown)


GoMati-MU (approved):

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3StorageFilter.scala
```diff
@@ -0,0 +1,32 @@
...
+  * Avoids reading objects that are in Glacier or Deep Archive storage classes.
+  * These objects are not immediately available and require a restore operation to be performed.
+  * This filter is used to avoid reading these objects and failing the task and thus the connector
+  */
+object AwsS3StorageFilter {
+
+  def filterOut(o: S3Object) =
+    o.storageClass() == ObjectStorageClass.GLACIER ||
+      o.storageClass() == ObjectStorageClass.DEEP_ARCHIVE ||
+      o.storageClass() == ObjectStorageClass.GLACIER_IR
```
davidsloan, line 31: Totally MINOR and STYLISTIC so feel free to ignore.  Would it be nicer to have a set of these `ObjectStorageClass`es and then have 

```
val invalidObjectStorageClasses = Set(GLACIER, DEEP_ARCHIVE, GLACIER_IR)

def filterOut(o: S3Object) = invalidStorageClasses.contains(o.storageClass)
```

davidsloan (approved):

kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/storage/AwsS3StorageFilter.scala
stheppi: done

PR #1220, opened 2024-05-13
LC-202 Backing up multiple topics at once

stheppi:
To back the topics up the KCQL statemet is

```
INSERT INTO bucket
SELECT * FROM `*`
...
```

When `*` is used the envelope setting is ignored.

This change allows for the `*` to be taken into account as a default if the given message topic is not found.

### DIFF
--- a/kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTaskAvroEnvelopeTest.scala
+++ b/kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTaskAvroEnvelopeTest.scala
@@ -84,7 +84,7 @@ class S3SinkTaskAvroEnvelopeTest
     record
   }
 
-  /*"S3SinkTask" should "write to avro format" in {
+  "S3SinkTask" should "write to avro format" in {
 
     testWritingAvro(
       (
@@ -93,7 +93,7 @@ class S3SinkTaskAvroEnvelopeTest
       ).asJava,
       "streamReactorBackups/myTopic/000000000001/000000000003_10001_10003.avro",
     )
-  }*/
+  }
 
   "S3SinkTask" should "write envelope when * is used as KCQL source" in {
 
--- a/kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTaskAvroEnvelopeTest.scala
+++ b/kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTaskAvroEnvelopeTest.scala
@@ -84,12 +84,23 @@ class S3SinkTaskAvroEnvelopeTest
     record
   }
 
-  "S3SinkTask" should "write to avro format using V1 format" in {
+  /*"S3SinkTask" should "write to avro format" in {
 
     testWritingAvro(
       (
         defaultProps +
           ("connect.s3.kcql" -> s"insert into $BucketName:$PrefixName select * from $TopicName STOREAS AVRO PROPERTIES('store.envelope'=true, 'padding.length.partition'='12', 'padding.length.offset'='12', '${FlushCount.entryName}'=3)")
+      ).asJava,
+      "streamReactorBackups/myTopic/000000000001/000000000003_10001_10003.avro",
+    )
+  }*/
+
+  "S3SinkTask" should "write envelope when * is used as KCQL source" in {
+
+    testWritingAvro(
+      (
+        defaultProps +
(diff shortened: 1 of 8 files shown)


kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTaskAvroEnvelopeTest.scala
```diff
@@ -84,7 +84,7 @@ class S3SinkTaskAvroEnvelopeTest
     record
   }
 
-  "S3SinkTask" should "write to avro format using V1 format" in {
+  /*"S3SinkTask" should "write to avro format" in {
```
davidsloan, line 87: Is this test case no longer valid?  Can it be removed?  Does the fact it needs to be commented out represent a breaking change that will need to be communicated to customers?

kafka-connect-aws-s3/src/test/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/S3SinkConfigDefBuilderTest.scala
```diff
@@ -229,6 +229,23 @@ class S3SinkConfigDefBuilderTest extends AnyFlatSpec with MockitoSugar with Matc
...
+    config.CloudSinkBucketOptions(connectorTaskId, S3SinkConfigDefBuilder(props)) match {
+      case Left(value) => fail(value.toString)
+      case Right(value) =>
+        value.map(_.dataStorage) should be(List(DataStorageSettings(envelope = true,
+                                                                    key      = true,
+                                                                    value    = true,
+                                                                    metadata = true,
+                                                                    headers  = true,
+        )))
+    }
```
davidsloan, line 246: The pattern match and `fail()` is not really necessary if you use scalatest's `EitherValues` trait.

```suggestion
    config.CloudSinkBucketOptions(connectorTaskId, S3SinkConfigDefBuilder(props)).value.map(_.dataStorage) should be(
      List(
        DataStorageSettings(
          envelope = true,
          key      = true,
          value    = true,
          metadata = true,
          headers  = true,
        )
      )
    )
```

kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/transformers/Transformers.scala
```diff
@@ -37,22 +38,29 @@ case class SequenceTransformer(transformers: Transformer*) extends Transformer {
...
-
     val transformersMap =
       bucketOptions
-        .filter(_.sourceTopic.nonEmpty)
         .foldLeft(Map.empty[Topic, Transformer]) {
           case (map, bo) =>
             if (bo.dataStorage.hasEnvelope) {
-              val topic = Topic(bo.sourceTopic.get)
+              //replace empty with `*`. This is because CloudSinkBucketOptions moves `*` to empty
+              val topic = Topic(bo.sourceTopic.getOrElse("*"))
```
davidsloan, line 63: Do you think it would be better if `CloudSinkBucketOptions` was configured differently so that the `*` was more intuitive?  Instead of a generic `Option` a custom type could be introduced.

I'm not suggesting you make any changes but perhaps there is some technical debt here if this tripped you up in the past, and it makes me think that there could be improvements made to make the configuration code more understandable?

davidsloan (approved):

kafka-connect-aws-s3/src/it/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTaskAvroEnvelopeTest.scala
stheppi: sorry, it is needed, I commented to test only the other test added. my mistake

kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/transformers/Transformers.scala
stheppi: yes, I think it is probably going to be clear a "default" option if enabled

kafka-connect-aws-s3/src/test/scala/io/lenses/streamreactor/connect/aws/s3/sink/config/S3SinkConfigDefBuilderTest.scala
davidsloan: Thanks for changing all those tests!

kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/transformers/Transformers.scala
davidsloan: I raised LC-204 on Jira to cover improvements here.

davidsloan (approved):

PR #1222, opened 2024-05-14
Azure Servicebus Connector

GoMati-MU:
#### Description

Initial version of Azure ServiceBus Connector. This is based on Aa[PR](https://github.com/lensesio/stream-reactor/pull/732/files#diff-078ec1a687c4a1d6531da6bf99acd3ec15484115758346472a8fd85f8ce69f1c) which is 3 years old but translated to java and cleaned up. It has some notable differences (e.g. it's no longer needed to construct `connectionString` as Azure provides it whole, we also don't create Administration client as it required highest privileges which could be a security issue). It also bumps up Lombok version and cleans up some areas (e.g. [val is now final var](https://projectlombok.org/features/val)).

Feedback is very welcome.

#### Testing

- [x] All major functionalities unit tests added.
- [x] manual local testing with actual Service Bus Topics
- [x] manual local testing with actual Service Bus Queues


java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceTask.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.kafka.connect.source.SourceTask;
+import org.apache.kafka.connect.storage.OffsetStorageReader;
+
+/**
+ * Implementation of {@link SourceTask} for Microsoft Azure EventHubs.
+ */
+@Slf4j
+public class AzureServiceBusSourceTask extends SourceTask {
+
+  private static final int RECORDS_QUEUE_DEFAULT_SIZE = 20;
```
GoMati-MU, line 36: shall this be configurable via connector props?

GoMati-MU:
The CVE is already an [issue on Microsoft side](https://github.com/jeremylong/DependencyCheck/issues/6100), but they don't seem keen to fixing it anytime soon

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnectorTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+    return propertyMap;
+  }
+
+  private String buildKcqlWithNecessaryProperties() {
+    final StringBuilder kcqlBuilder = new StringBuilder("INSERT INTO output SELECT * FROM input ");
+    kcqlBuilder.append("PROPERTIES(");
+
+    for (ServiceBusKcqlProperties property : ServiceBusKcqlProperties.values()) {
+      kcqlBuilder.append("'" + property.getPropertyName() + "'='" + property.getPropertyName() + "',");
```
davidsloan, line 128: Let's not mix StringBuilders and String concatenation, picking one or the other and sticking to it would be better.

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/ServiceBusReceiverFacadeTest.java
```diff
@@ -0,0 +1,126 @@
...
+  void checkReceiverInitializationAndReceiverIdForQueue() {
+    //given
+    String queueType = "QUEUE";
+    Map<String, String> propertiesMap = mock(Map.class);
+    when(propertiesMap.get(ServiceBusKcqlProperties.SERVICE_BUS_TYPE.getPropertyName())).thenReturn(queueType);
+
+    String inputBusName = "INPUT";
+    Kcql kcql = mock(Kcql.class);
+    when(kcql.getSource()).thenReturn(inputBusName);
+    when(kcql.getProperties()).thenReturn(propertiesMap);
```
davidsloan, line 50: I might split these up into a simple utility method like 

```

  private static Kcql mockKcql(String inputBusName, Map<String, String> propertiesMap) {
    Kcql kcql = mock(Kcql.class);
    when(kcql.getSource()).thenReturn(inputBusName);
    when(kcql.getProperties()).thenReturn(propertiesMap);
    return kcql;
  }
```

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ExecutorService;
+import org.apache.kafka.connect.source.SourceRecord;
+import org.apache.kafka.connect.storage.OffsetStorageReader;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class TaskToReceiverBridgeTest {
+
+  Map<String, String> properties = new HashMap<>();
```
davidsloan, line 39: This field is never used.

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
```diff
@@ -0,0 +1,130 @@
...
+  ArrayBlockingQueue<ServiceBusMessageHolder> blockingQueue;
+  OffsetStorageReader offsetReader;
+
+  TaskToReceiverBridge testObj;
+
+  @BeforeEach
+  void setUp() {
+    blockingQueue = mock(ArrayBlockingQueue.class);
+    offsetReader = mock(OffsetStorageReader.class);
```
davidsloan, line 48: This causes an unsafe cast warning.

My suggestion is to use @ExtendWith(MockitoJUnitExtension.class) and then you can annotate the fields as `@Mock`.  Then you can probably remove the `setUp` method.

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
```diff
@@ -0,0 +1,130 @@
...
+  @Test
+  void closeReceiversShouldCloseAllReceivers() {
+    //given
+    ExecutorService executorService = mock(ExecutorService.class);
+    String receiverId1 = "RECEIVER1";
+    String receiverId2 = "RECEIVER2";
+    ServiceBusReceiverFacade receiver1 = mock(ServiceBusReceiverFacade.class);
+    ServiceBusReceiverFacade receiver2 = mock(ServiceBusReceiverFacade.class);
+    when(receiver1.getReceiverId()).thenReturn(receiverId1);
+    when(receiver2.getReceiverId()).thenReturn(receiverId2);
```
davidsloan, line 60: How about extracting a method to mock a receiver?

```

    String receiverId1 = "RECEIVER1";
    ServiceBusReceiverFacade receiver1 = mockReceiver(receiverId1);

    String receiverId2 = "RECEIVER2";
    ServiceBusReceiverFacade receiver2 = mockReceiver(receiverId2);
```

Then afterwards `receiverId1` and `receiverId2` could be made constants as they are being used in other tests too.

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
```diff
@@ -0,0 +1,130 @@
...
+      sourceRecordBlockingQueue.put(mockedRecord);
+    }
+
+    //when
+    testObj = new TaskToReceiverBridge(sourceRecordBlockingQueue, offsetReader, receivers, executorService);
+    List<SourceRecord> polled = testObj.poll();
+
+    //then
+    assertThat(polled).hasSize(arrayBlockingQueueCapacity);
+    assertThat(polled).containsExactlyElementsOf(allSourceRecords);
```
davidsloan, line 106: This should really be joined together in a single assertion chain:

`    assertThat(polled).hasSize(arrayBlockingQueueCapacity).containsExactlyElementsOf(allSourceRecords);
`

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
```diff
@@ -0,0 +1,130 @@
...
+    for (int i = 0; i < arrayBlockingQueueCapacity; i++) {
+      ServiceBusMessageHolder mockedRecord = mock(ServiceBusMessageHolder.class);
+      ServiceBusReceivedMessage busReceivedMessage = mock(ServiceBusReceivedMessage.class);
+      SourceRecord sourceRecord = mock(SourceRecord.class);
+      when(mockedRecord.getTranslatedRecord()).thenReturn(sourceRecord);
+      when(mockedRecord.getOriginalRecord()).thenReturn(busReceivedMessage);
+      when(busReceivedMessage.getMessageId()).thenReturn(String.format(messageIdTemplate, i));
+      allSourceRecords.add(sourceRecord);
+      sourceRecordBlockingQueue.put(mockedRecord);
+    }
```
davidsloan, line 98: A good rule of thumb when working with mocks, is to try and keep the mock creation and behaviour together.

From your code I can tweak it a little to group the mock definition and behaviours and make it a little easier to read:

Also using an IntStream and then a collector can be a good more functional alternative to building an ArrayList incrementally.

```

    List<SourceRecord> allSourceRecords = IntStream.range(0, arrayBlockingQueueCapacity).mapToObj(i -> {

      SourceRecord sourceRecord = mock(SourceRecord.class);

      ServiceBusReceivedMessage busReceivedMessage = mock(ServiceBusReceivedMessage.class);
      when(busReceivedMessage.getMessageId()).thenReturn(String.format(messageIdTemplate, i));

      ServiceBusMessageHolder mockedRecord = mock(ServiceBusMessageHolder.class);
      when(mockedRecord.getTranslatedRecord()).thenReturn(sourceRecord);
      when(mockedRecord.getOriginalRecord()).thenReturn(busReceivedMessage);
      try {
        sourceRecordBlockingQueue.put(mockedRecord);
      } catch (InterruptedException e) {
        throw new RuntimeException(e);
      }
      return sourceRecord;
    }).collect(Collectors.toList());
```

From this point you could start extracting out methods for reusability/clarity.

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/util/KcqlConfigBusMapperTest.java
```diff
@@ -0,0 +1,203 @@
...
+    StringBuilder fullKcql = new StringBuilder();
+
+    for (int i = 0; i < numberOfMappings; i++) {
+      String newInput = "INPUT" + i;
+      String newOutput = "OUTPUT" + i;
+
+      inputs.add(i, newInput);
+      outputs.add(i, newOutput);
+      fullKcql.append(String.format(kcqlTemplate, newOutput, newInput));
+      fullKcql.append(" " + createNecessaryPropertiesPart() + ";");
```
davidsloan, line 52: As above, let's not mix StringBuilder and string concatenation.

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/util/KcqlConfigBusMapperTest.java
```diff
@@ -0,0 +1,203 @@
...
+            + "dots and has to start with number or letter with max size of 160";
+    String outputErrorMessage =
+        "Output topic OUTPUT*_, name is not correctly specified: "
+            + "It can contain only letters, numbers and hyphens, underscores and "
+            + "dots and has to start with number or letter with max size of 160";
+
+    //when
+    mapInputToOutputAssertingExceptionWithSpecificMessage(illegalInputKcql, inputErrorMessage);
+
+    mapInputToOutputAssertingExceptionWithSpecificMessage(illegalOutputKcql, outputErrorMessage);
```
davidsloan, line 89: Good reuse 👍

davidsloan (changes requested):
Good test coverage, but a few things to think about structure wise.  Some of the things I've highlighted are stylistic - we tend to favour a more functional style.

Please reach out if I can help you with anything or if you would like to discuss in more detail.

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
```diff
@@ -0,0 +1,136 @@
...
+import org.apache.kafka.connect.source.SourceRecord;
+import org.apache.kafka.connect.storage.OffsetStorageReader;
+
+/**
+ * Bridge between Receivers and Connector's Task for Azure Service Bus.
+ */
+@Slf4j
+public class TaskToReceiverBridge {
+
+  private static final int INITIAL_RECORDS_TO_COMMIT_SIZE = 500;
```
davidsloan, line 40: Should this be configurable via KCQL?

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnector.java
```diff
@@ -0,0 +1,80 @@
...
+    return AzureServiceBusSourceTask.class;
+  }
+
+  @Override
+  public List<Map<String, String>> taskConfigs(int maxTasks) {
+    log.info("Setting task configurations for {} workers.", maxTasks);
+    List<Map<String, String>> taskConfigs = new ArrayList<>(maxTasks);
+
+    IntStream.range(0, maxTasks).forEach(task -> taskConfigs.add(configProperties));
+    return taskConfigs;
```
davidsloan, line 63: This question is for @stheppi and @andrewstevenson .

Should the KCQLs be split between the multiple tasks like in [the original PR](https://github.com/lensesio/stream-reactor/pull/732/files#diff-078ec1a687c4a1d6531da6bf99acd3ec15484115758346472a8fd85f8ce69f1cR44-R65)?



I've asked @GoMati-MU  about this and he was kind enough to point me to the acceptance criteria on the [jira task](https://landoop.atlassian.net/browse/LC-181) which states:

> - the parallelism is matching the task.max . this means that each task connects to all the provided topics/queue

Just wanted to double check 🙇

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
GoMati-MU: I don't think so, as it'll resize if needed (it's only INITIAL). 

If you think otherwise, then sure.

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/util/KcqlConfigBusMapperTest.java
GoMati-MU: yup!

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
GoMati-MU: I can get along on grouping, using Stream or for-loop gives the same results.

I wonder however, why try-catch here instead of just telling that test can throw Exception? It's not a production code.

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
GoMati-MU: okay

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
GoMati-MU: Done

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
GoMati-MU: But underneath it would've done the same thing, wouldn't it? It'd just have runner on top of it.

It's just how Mockito works 🙂

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
GoMati-MU: Removed, thanks!

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/ServiceBusReceiverFacadeTest.java
GoMati-MU: okay

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnectorTest.java
GoMati-MU: Agreed!

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/StringUtils.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.lenses.streamreactor.common.util;
+
+public class StringUtils {
+
+  public static boolean isBlank(String str) {
+    return str == null || str.trim().isEmpty();
+  }
```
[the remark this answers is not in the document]
GoMati-MU: But wouldn't I be then testing Java API really? (I tend to avoid that)

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/util/KcqlConfigBusMapper.java
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * This method parses KCQL statements and fetches input and output topics checking against regex for invalid
+   * topic names in input and output as well as if mappings contain necessary KCQL properties.
+   * 
+   * @param kcqlString string to parse
+   * @return map of input to output topic names
+   */
+  public static Map<String, String> mapInputToOutputsFromConfig(String kcqlString) {
+    List<Kcql> kcqls = Kcql.parseMultiple(kcqlString);
```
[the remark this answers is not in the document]
GoMati-MU: It turns out I wasn't really using KCQL mapper as I intended to in the first place.

Thanks a lot, it should be nicer now!

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
```diff
@@ -0,0 +1,136 @@
...
+   * 
+   * @param properties          all properties for {@link org.apache.kafka.connect.source.SourceTask}.
+   * @param recordsQueue        records queue used to store received messages.
+   * @param offsetStorageReader offset storage reader from Task.
+   */
+  public TaskToReceiverBridge(Map<String, String> properties,
+      BlockingQueue<ServiceBusMessageHolder> recordsQueue, OffsetStorageReader offsetStorageReader) {
+    this.recordsQueue = recordsQueue;
+    this.offsetProvider = new ServiceBusPartitionOffsetProvider(offsetStorageReader);
+    String connectionString = properties.get(AzureServiceBusConfigConstants.CONNECTION_STRING);
```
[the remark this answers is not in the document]
GoMati-MU: It is interesting idea, but I'm not sure I see the benefits clearly. The configuration will still be passed around, it's just gonna be form (POJO) that's different, isn't it?

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceTask.java
```diff
@@ -0,0 +1,88 @@
...
+  public String version() {
+    return jarManifest.getVersion();
+  }
+
+  @Override
+  public void start(Map<String, String> props) {
+    OffsetStorageReader offsetStorageReader =
+        ofNullable(this.context).flatMap(
+            context -> ofNullable(context.offsetStorageReader())).orElseThrow();
+    AzureServiceBusSourceConfig azureServiceBusSourceConfig = new AzureServiceBusSourceConfig(props);
```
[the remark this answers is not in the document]
GoMati-MU: It is now :) Thanks for head-up though!

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnector.java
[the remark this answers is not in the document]
GoMati-MU: I think it does the same in this context (`limit` has some downsides for parallel, but nothing bad here, it also creates additional unnecessary object, but I'm just being picky).

I just think my approach is easier to read (may be just an opinion)

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnector.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.kafka.connect.source.SourceConnector;
+
+/**
+ * Implementation of {@link SourceConnector} for Microsoft Azure EventHubs.
+ */
+@Slf4j
+public class AzureServiceBusSourceConnector extends SourceConnector {
+
+  private final JarManifest jarManifest =
+      new JarManifest(getClass().getProtectionDomain().getCodeSource().getLocation());
```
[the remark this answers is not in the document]
GoMati-MU: The reason why I would be kind-of against that is the _need-to-know_ principle. JarManifest doesn't really care about the class, all it needs is `URL` or `JarFile`.

It's not necessarily wrong, I can see the intention, but it's not without downsides IMO.

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/mapping/ServiceBusValueSchemaField.java
```diff
@@ -0,0 +1,114 @@
...
+    ALL_FIELDS.add(REPLY_TO_SESSION_ID);
+    ALL_FIELDS.add(DEAD_LETTER_SOURCE);
+    ALL_FIELDS.add(TIME_TO_LIVE);
+    ALL_FIELDS.add(LOCKED_UNTIL_UTC);
+    ALL_FIELDS.add(SEQUENCE_NUMBER);
+    ALL_FIELDS.add(SESSION_ID);
+    ALL_FIELDS.add(LOCK_TOKEN);
+    ALL_FIELDS.add(MESSAGE_BODY);
+    ALL_FIELDS.add(GET_TO);
+  }
```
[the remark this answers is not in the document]
GoMati-MU: yup!

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/mapping/ServiceBusValueSchemaField.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.List;
+import org.apache.kafka.connect.data.Field;
+import org.apache.kafka.connect.data.Schema;
+
+/**
+ * Class that represents Kafka Connects {@link Schema} for values from Service Bus.
+ */
+public class ServiceBusValueSchemaField {
+
+  private static int NUMBER_OF_FIELDS = 17;
```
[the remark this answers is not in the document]
GoMati-MU: For sure!

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/mapping/ServiceBusToSourceRecordMapper.java
```diff
@@ -0,0 +1,99 @@
...
+    targetObject.put(REPLY_TO_SESSION_ID, serviceBusMessage.getReplyToSessionId());
+    targetObject.put(DEAD_LETTER_SOURCE, serviceBusMessage.getDeadLetterSource());
+    targetObject.put(TIME_TO_LIVE, serviceBusMessage.getTimeToLive().toMillis());
+    targetObject.put(LOCKED_UNTIL_UTC, serviceBusMessage.getLockedUntil().toEpochSecond());
+    targetObject.put(SEQUENCE_NUMBER, serviceBusMessage.getSequenceNumber());
+    targetObject.put(SESSION_ID, serviceBusMessage.getSessionId());
+    targetObject.put(LOCK_TOKEN, serviceBusMessage.getLockToken());
+    targetObject.put(MESSAGE_BODY, serviceBusMessage.getBody().toBytes());
+    targetObject.put(GET_TO, serviceBusMessage.getTo());
+  }
```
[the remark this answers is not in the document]
GoMati-MU: Makes sense, changed

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/mapping/AzureServiceBusSourceRecord.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Implementation of {@link SourceRecord} for Microsoft Azure EventHubs.
+ */
+@ToString(callSuper = true)
+public class AzureServiceBusSourceRecord extends SourceRecord {
+
+  public AzureServiceBusSourceRecord(Map<String, ?> sourcePartition, Map<String, ?> sourceOffset,
+      String topic, Integer partition, Schema keySchema, Object key,
+      Schema valueSchema, Object value, Long timestamp, Iterable<Header> headers) {
+    super(sourcePartition, sourceOffset, topic, partition, keySchema, key, valueSchema, value, timestamp, headers);
```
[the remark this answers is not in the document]
GoMati-MU: Makes sense, changed

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceTask.java
GoMati-MU: Good idea, added!

davidsloan: It was your idea 😁

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.junit.jupiter.api.Test;
+
+class TaskToReceiverBridgeTest {
+
+  private static final String RECEIVER_ID_1 = "RECEIVER1";
+  private static final String RECEIVER_ID_2 = "RECEIVER2";
+  ArrayBlockingQueue<ServiceBusMessageHolder> blockingQueue;
+  OffsetStorageReader offsetReader;
+
+  TaskToReceiverBridge testObj;
```
davidsloan, line 44: these can be locked down to `private`

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
davidsloan: This is discipline to avoid the errors/violations/problems appearing in the IDE/static analysis tools, so we can focus on the real problems without adding extra noise.

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
davidsloan: @GoMati-MU this is from [your code.](
https://github.com/lensesio/stream-reactor/blob/634953a5877c38377ac9975ebe9e3ba1752a948a/java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java#L94)  I suppose I commented on the wrong method, but the amount of duplication here needs to be resolved.

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
davidsloan: In all honesty I'm not sure what it is - can we add some javadoc around it to explain in more detail?

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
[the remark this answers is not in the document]
davidsloan: Single Responsibility Principle

We should have code that handles KCQL parsing and configuration concerns.  These should not be the same code as the connector logic.  Also it would reduce the duplication I identified around parsing the KCQL twice.

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
```diff
@@ -0,0 +1,156 @@
...
+   * Polls for Consumer Records from the queue.
+   * 
+   * @return List of {@link SourceRecord} or empty list if no new messages received.
+   */
+  public List<SourceRecord> poll() {
+    List<ServiceBusMessageHolder> recordsFromQueue = new ArrayList<>(recordsQueue.size());
+
+    recordsQueue.drainTo(recordsFromQueue);
+
+    List<SourceRecord> recordsToSend =
```
davidsloan, line 131: unnecessary variable assignment

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * Creates Bridge between Receivers and Connector's Task for Azure Service Bus.
+   * 
+   * @param properties          all properties for {@link org.apache.kafka.connect.source.SourceTask}.
+   * @param recordsQueue        records queue used to store received messages.
+   * @param offsetStorageReader offset storage reader from Task.
+   */
+  public TaskToReceiverBridge(Map<String, String> properties,
+      BlockingQueue<ServiceBusMessageHolder> recordsQueue, OffsetStorageReader offsetStorageReader) {
```
davidsloan, line 56: Is this queue unbounded?  Will ServiceBus keep sending messages until we run out of memory if there is a delay in processing or polling?  What is the mechanism to handle backpressure?

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
```diff
@@ -0,0 +1,156 @@
...
+    this.receivers = receivers;
+    recordsToCommitMap = new ConcurrentHashMap<>(INITIAL_RECORDS_TO_COMMIT_SIZE);
+  }
+
+  private void initiateReceivers(BlockingQueue<ServiceBusMessageHolder> recordsQueue,
+      List<Kcql> kcqls, String connectionString) {
+    receivers = new ConcurrentHashMap<>(kcqls.size());
+
+    kcqls.stream().forEach(kcql -> {
+      String receiverId = ServiceBusReceiverFacade.class.getSimpleName() + UUID.randomUUID();
```
davidsloan, line 109: `ServiceBusReceiverFacade.class.getSimpleName()`
this should be extracted to a static final constant at the class level as it is (hopefully!) not going to change during runtime(!)

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
```diff
@@ -0,0 +1,156 @@
...
+  private final Map<String, ServiceBusMessageHolder> recordsToCommitMap;
+
+  /**
+   * Creates Bridge between Receivers and Connector's Task for Azure Service Bus.
+   * 
+   * @param properties          all properties for {@link org.apache.kafka.connect.source.SourceTask}.
+   * @param recordsQueue        records queue used to store received messages.
+   * @param offsetStorageReader offset storage reader from Task.
+   */
+  public TaskToReceiverBridge(Map<String, String> properties,
```
davidsloan, line 55: This constructor is never used.

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * Bridge between Receivers and Connector's Task for Azure Service Bus.
+ */
+@Slf4j
+public class TaskToReceiverBridge {
+
+  private static final int INITIAL_RECORDS_TO_COMMIT_SIZE = 500;
+  private final BlockingQueue<ServiceBusMessageHolder> recordsQueue;
+  private final ServiceBusPartitionOffsetProvider offsetProvider;
```
davidsloan, line 43: offsetProvider is written but never accessed

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
```diff
@@ -0,0 +1,156 @@
...
+   * @param kcqls               list of KCQLs to initiate Receivers for.
+   * @param recordsQueue        records queue used to store received messages.
+   * @param offsetStorageReader offset storage reader from Task.
+   */
+  public TaskToReceiverBridge(String connectionString, List<Kcql> kcqls,
+      BlockingQueue<ServiceBusMessageHolder> recordsQueue, OffsetStorageReader offsetStorageReader) {
+    this.recordsQueue = recordsQueue;
+    this.offsetProvider = new ServiceBusPartitionOffsetProvider(offsetStorageReader);
+    sentMessagesExecutors = Executors.newFixedThreadPool(kcqls.size() * 2);
+    recordsToCommitMap = new ConcurrentHashMap<>(INITIAL_RECORDS_TO_COMMIT_SIZE);
```
davidsloan, line 81: Let's think about this class from a dependency injection point of view - instead of having 2 constructors, with a separate one that is only accessed in test code, it might be beneficial to have a single constructor used everywhere .

I would definitely advocate moving object creation responsibilities higher up in the code hierarchy, for example it could all happen in the `SourceTask`/`SourceConnector` only.

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
```diff
@@ -0,0 +1,156 @@
...
+    this.sentMessagesExecutors = sentMessagesExecutors;
+    this.receivers = receivers;
+    recordsToCommitMap = new ConcurrentHashMap<>(INITIAL_RECORDS_TO_COMMIT_SIZE);
+  }
+
+  private void initiateReceivers(BlockingQueue<ServiceBusMessageHolder> recordsQueue,
+      List<Kcql> kcqls, String connectionString) {
+    receivers = new ConcurrentHashMap<>(kcqls.size());
+
+    kcqls.stream().forEach(kcql -> {
```
davidsloan, line 108: You can just do `kcqls.forEach(kcql -> {...`, on this occasion you don't need to stream 😁

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/ServiceBusPartitionOffsetProvider.java
```diff
@@ -0,0 +1,92 @@
...
+      this.put(PARTITION_KEY, partition);
+    }
+
+    public String getTopic() {
+      return get(BUS_KEY);
+    }
+
+    public Integer getPartition() {
+      return Integer.valueOf(get(PARTITION_KEY));
+    }
```
davidsloan, line 75: These getters are never used, what are they for?

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/ServiceBusReceiverFacade.java
```diff
@@ -0,0 +1,157 @@
...
+      try {
+        SourceRecord sourceRecord = mapSingleServiceBusMessage(message, outputTopic, partitionKey, offsetMarker);
+        ServiceBusMessageHolder serviceBusMessageHolder =
+            new ServiceBusMessageHolder(message, sourceRecord, receiverId);
+        boolean offer = false;
+        while (!offer) {
+          offer = recordsQueue.offer(serviceBusMessageHolder, FIVE_SECONDS_TIMEOUT, TimeUnit.SECONDS);
+        }
+      } catch (InterruptedException e) {
+        log.info("{} has been interrupted on offering", receiverId);
```
davidsloan, line 152: Should we consider additional handling for the InterruptedException beyond logging? For example, should we retry the operation, perform specific cleanup actions, or propagate the exception to maintain the application's interrupted state?

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/StringUtils.java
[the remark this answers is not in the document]
davidsloan: I don't follow.  If this is part of the JDK then why do we need the class?


Here's a sample implementation for you to consider:

```
package io.lenses.streamreactor.common.util;

import static org.junit.jupiter.api.Assertions.assertFalse;
import static org.junit.jupiter.api.Assertions.assertTrue;

import org.junit.jupiter.api.Test;

class StringUtilsTest {

    @Test
    void testIsBlankWithNull() {
        assertTrue(StringUtils.isBlank(null), "null should be considered blank");
    }

    @Test
    void testIsBlankWithEmptyString() {
        assertTrue(StringUtils.isBlank(""), "Empty string should be considered blank");
    }

    @Test
    void testIsBlankWithWhitespace() {
        assertTrue(StringUtils.isBlank("   "), "String with only whitespace should be considered blank");
    }

    @Test
    void testIsBlankWithNonBlankString() {
        assertFalse(StringUtils.isBlank("abc"), "Non-blank string should not be considered blank");
    }

    @Test
    void testIsBlankWithStringContainingWhitespace() {
        assertFalse(StringUtils.isBlank(" abc "), "String with non-whitespace characters should not be considered blank");
    }
}
```

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/config/AzureServiceBusSourceConfig.java
```diff
@@ -0,0 +1,87 @@
...
+                AzureServiceBusConfigConstants.TASK_RECORDS_QUEUE_SIZE
+            ).define(AzureServiceBusConfigConstants.CONNECTION_STRING,
+                Type.STRING,
+                Importance.HIGH,
+                AzureServiceBusConfigConstants.CONNECTION_STRING_DOC,
+                CONNECTION_GROUP,
+                2,
+                ConfigDef.Width.LONG,
+                AzureServiceBusConfigConstants.CONNECTION_STRING
+            ).define(AzureServiceBusConfigConstants.KCQL_CONFIG,
```
davidsloan, line 67: I wanted to point you to [this](https://github.com/lensesio/stream-reactor/blob/26d8f2652e8197a7aa1cc5cc7c4311194414969b/kafka-connect-azure-datalake/src/main/scala/io/lenses/streamreactor/connect/datalake/auth/DatalakeClientCreator.scala#L41)

Customers won't always be able to use connection string for connecting to azure, especially if they are running on an Azure instance in which they will have default chain from the instance profile.  I believe the original PR was more complete in this respect?

But if we can we should identify code we can reuse here and move it into the Java repo so we are not inventing the wheel and duplicating throughout the codebase.  I did similar recently for GCP cloud connectors, so can help with this if you would like.

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnector.java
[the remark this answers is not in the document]
GoMati-MU: If it's a feature around `JarManifest` I guess it should be a separate discussion, as `JarManifest` is not even part of this connector, right? I think we can leave it here if reasoning seems right for now.

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnector.java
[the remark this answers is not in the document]
GoMati-MU: As said: I think this is less readable, not to mention both are using functional (stream) approach. I think we should talk it through really.

But if we want to be nitpicky, I will change it to suggested approach.

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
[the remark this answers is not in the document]
GoMati-MU: It's about having single responsibility (being bridge between facade and task), not reducing the dependencies (constructor arguments). But I just moved logic around handling of connectionString and KCQL out of this class as you wished.

java-connectors/kafka-connect-common/src/main/java/io/lenses/streamreactor/common/util/StringUtils.java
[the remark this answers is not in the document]
GoMati-MU: Added, thank you!

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
GoMati-MU: Changed, but it's a bit like sweeping not-really-existent problem under the rug

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
GoMati-MU: We're running in circles :) This is *my code* because I used your example.

But I know the answer: because we use stream there's no other way to deal with that exception, with `for` it'd be way more lenient.

Test has been updated

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
GoMati-MU: I mean: it's used as an initial size to initialize `recordsToCommit` in constructors:
`recordsToCommitMap = new ConcurrentHashMap<>(INITIAL_RECORDS_TO_COMMIT_SIZE);`

Maybe I can even give it a better name, but what would it be?

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/mapping/ServiceBusValueSchemaField.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.List;
+import org.apache.kafka.connect.data.Field;
+import org.apache.kafka.connect.data.Schema;
+
+/**
+ * Class that represents Kafka Connects {@link Schema} for values from Service Bus.
+ */
+public class ServiceBusValueSchemaField {
+
+  private static final int NUMBER_OF_FIELDS = 17;
```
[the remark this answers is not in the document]
GoMati-MU: Removed, thanks!

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/mapping/ServiceBusValueSchemaField.java
```diff
@@ -0,0 +1,110 @@
...
+          GET_TO);
+
+  private ServiceBusValueSchemaField() {
+  }
+
+  public static List<Field> getAllFields() {
+    return ALL_FIELDS;
+  }
+
+  private class SchemaFieldConstants {
```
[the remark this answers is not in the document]
GoMati-MU: Changed to static, thanks!

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/mapping/ServiceBusValueSchemaField.java
```diff
@@ -0,0 +1,110 @@
...
+  public static final Field SEQUENCE_NUMBER =
+      new Field(SchemaFieldConstants.SEQUENCE_NUMBER, 12, Schema.OPTIONAL_INT64_SCHEMA);
+  public static final Field SESSION_ID =
+      new Field(SchemaFieldConstants.SESSION_ID, 13, Schema.OPTIONAL_STRING_SCHEMA);
+  public static final Field LOCK_TOKEN =
+      new Field(SchemaFieldConstants.LOCK_TOKEN, 14, Schema.OPTIONAL_STRING_SCHEMA);
+  public static final Field MESSAGE_BODY =
+      new Field(SchemaFieldConstants.MESSAGE_BODY, 15, Schema.BYTES_SCHEMA);
+  public static final Field GET_TO =
+      new Field(SchemaFieldConstants.GET_TO, 16, Schema.OPTIONAL_STRING_SCHEMA);
```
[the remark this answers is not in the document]
GoMati-MU: Agreed, changed.

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
GoMati-MU: Done

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
GoMati-MU: Changed

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
GoMati-MU: It is bounded (and now size is configurable), we initialize it in `AzureServiceBusSourceTask`. If by backpressure we mean committing ServiceBus offsets after message's been delivered, we just use `commitRecord` method in the `AzureServiceBusSourceTask`

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
GoMati-MU: Good spot, removed!

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
GoMati-MU: Removed

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
GoMati-MU: Then the branching will still have to happen, but will be in `SourceTask` which essentially doesn't have to know about Bridge's implementation details (that it initializes Facades).

I'd advise leaving it as-is.

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
GoMati-MU: Good spot, removed!

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
GoMati-MU: Removed

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
GoMati-MU: Then the branching will still have to happen, but will be in `SourceTask` which essentially doesn't have to know about Bridge's implementation details (that it initializes Facades).

I'd advise leaving it as-is.

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/config/AzureServiceBusSourceConfig.java
GoMati-MU: There are two ways described in official Service Bus docs: one being **Passwordless**, second being **Connection String**. Since the first one will work using specifically-installed Azure commandline and our connectors are being ran inside docker container, it would require us to include Azure commandline and ways to work with it within Lenses Box (which will no doubt be huge amounts of work).

Connection String already proved to work for our clients in other Azure connectors, so that's why I've chosen to implement this path.

My take on it is: unless we get a specific requirement to support another type of authentication, let's just use what we have (as this is another feature that we don't know if ever will be used yet)

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/ServiceBusReceiverFacade.java
GoMati-MU: Good question! For now:
1. If record won't reach recordsQueue it won't be sent to Kafka
2. Therefore it won't be committed in Service Bus (the logic to commit it won't work), which will mean we'll have to process it again anyway.

However, I still consider this a valid point. Log probably should be more visible, so I will definitely bump it's level to WARN (we can think if ERROR too) and maybe we can also consider timeout to be configurable?

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/ServiceBusPartitionOffsetProvider.java
GoMati-MU: Removed, thanks!

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
GoMati-MU: Good spot, changed!

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
davidsloan: There's no branching from what I can see? Please can you explain?

I've committed a [small example of how this refactoring could be achieved](#1247 ).

The creation code can be extracted out to a `ServiceBusReceiverMapper`, which is responsible for the creation logic/wiring.  Mostly this is what `createRecievers` was doing.
The benefits of this:

- `TaskToReceiverBridge` only has a single constructor.
- `TaskToReceiverBridge` no longer has to create the receivers, single responsibility principal.  This now falls to `ServiceBusReceiverMapper`.
- `TaskToReceiverBridge` no longer has to know about the connection string.
- Test and production code can now share the same constructor.  There's no longer a special 'test' constructor.
- All fields of `TaskToReceiverBridge` can now be made final.
- Object creation can be pushed up the stack as previously described into `AzureServiceBusSourceTask`

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/util/KcqlConfigBusMapper.java
```diff
@@ -0,0 +1,124 @@
...
+            String.format("Following non-optional properties missing in KCQL: %s", missingPropertiesError));
+      }
+
+      checkForValidPropertyValues(kcql.getProperties());
+
+      inputToOutputTopics.put(inputTopic, outputTopic);
+    }
+
+    return kcqls;
+  }
```
davidsloan, line 97: Could we split the validation logic and mapping logic into separate concerns?

Something like (this isn't a complete implementation):

```suggestion
  public static List<Kcql> mapKcqlsFromConfig(String kcqlString) {
    List<Kcql> kcqls = Kcql.parseMultiple(kcqlString);

    Map<String, String> inputToOutputTopics = kcqls.stream()
            .collect(Collectors.toMap(Kcql::getSource, Kcql::getTarget));

    validateMappings(inputToOutputTopics);

    return kcqls;
  }
```

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnector.java
```diff
@@ -0,0 +1,83 @@
...
+  @Override
+  public void start(Map<String, String> props) {
+    parseAndValidateConfigs(props);
+    configProperties = props;
+    printAsciiHeader(jarManifest, "/azure-servicebus-ascii.txt");
+  }
+
+  private void parseAndValidateConfigs(Map<String, String> props) {
+    new AzureServiceBusSourceConfig(props);
+    KcqlConfigBusMapper.mapKcqlsFromConfig(props.get(AzureServiceBusConfigConstants.KCQL_CONFIG));
```
davidsloan, line 52: Here we're relying on the fact that these two methods will throw exceptions if the configurations supplied are not correct.  It's easy to miss this reading the code.

How can we make this more obvious?

Wouldn't making sure each of these throw a specific checked exception that we would be forced to handle be the Java way of signalling this?

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
davidsloan: Very cool! 😎

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnectorTest.java
```diff
@@ -0,0 +1,121 @@
...
+    assertThat(taskConfigs.get(1)).isEqualTo(validProperties);
+  }
+
+  @Test
+  void configShouldReturnConfigClass() {
+    //when
+    ConfigDef configDef = testObj.config();
+
+    //then
+    AzureServiceBusSourceConfig.getConfigDefinition().equals(configDef);
```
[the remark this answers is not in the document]
GoMati-MU: Good spot, it now has assertion it should've had in the first place.

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnectorTest.java
```diff
@@ -0,0 +1,121 @@
...
+    verify(jarManifest).buildManifestString();
+  }
+
+  @Test
+  void taskClassShouldReturnTaskClass() {
+    //when
+    Class<? extends Task> taskClass = testObj.taskClass();
+
+    //then
+    AzureServiceBusSourceTask.class.equals(taskClass);
```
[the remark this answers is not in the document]
GoMati-MU: Good spot, it now has assertion it should've had in the first place.

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnectorTest.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  private Map<String, String> buildValidProperties() {
+    HashMap<String, String> propertyMap = new HashMap<>();
+
+    propertyMap.put(AzureServiceBusConfigConstants.CONNECTOR_NAME, CONFIG_CONNECTOR_NAME);
+    propertyMap.put(AzureServiceBusConfigConstants.CONNECTION_STRING, EMPTY_STRING);
+    propertyMap.put(AzureServiceBusConfigConstants.KCQL_CONFIG, VALID_KCQL);
+
+    return propertyMap;
```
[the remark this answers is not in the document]
GoMati-MU: Changed as I have nothing against that, but what difference does it make in unit tests utility method? I mean: in this context, does it change anything?

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/mapping/ServiceBusPartitionOffsetProviderTest.java
```diff
@@ -0,0 +1,83 @@
...
+    when(offsetStorageReader.offset(any(AzureServiceBusPartitionKey.class))).thenReturn(null);
+
+    //when
+    Optional<AzureServiceBusOffsetMarker> offset =
+        testObj.getOffset(new AzureServiceBusPartitionKey(someTopic, partition));
+
+    //then
+    verify(offsetStorageReader).offset(any(AzureServiceBusPartitionKey.class));
+    assertThat(offset).isNotNull();
+    assertThat(offset.isEmpty()).isTrue();
```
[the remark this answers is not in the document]
GoMati-MU: Changed

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/mapping/ServiceBusPartitionOffsetProviderTest.java
```diff
@@ -0,0 +1,83 @@
...
+    when(offsetStorageReader.offset(azureServiceBusPartitionKey)).thenReturn(offsetMap);
+
+    //when
+    Optional<AzureServiceBusOffsetMarker> offset =
+        testObj.getOffset(azureServiceBusPartitionKey);
+
+    //then
+    verify(offsetStorageReader).offset(any(AzureServiceBusPartitionKey.class));
+    assertThat(offset).isNotNull();
+    assertThat(offset.isEmpty()).isFalse();
```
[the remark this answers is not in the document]
GoMati-MU: Changed

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/mapping/ServiceBusPartitionOffsetProviderTest.java
```diff
@@ -0,0 +1,83 @@
...
+    assertThat(offset.get().getOffsetValue()).isEqualTo(exampleOffset);
```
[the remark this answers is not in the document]
GoMati-MU: Agreed, thanks!

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/util/ServiceBusKcqlPropertiesTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+  @Test
+  void getNecessaryProperties() {
+    //when
+    List<ServiceBusKcqlProperties> necessaryProperties = ServiceBusKcqlProperties.getNecessaryProperties();
+
+    //then
+    assertThat(necessaryProperties).hasSize(1);
+    assertThat(necessaryProperties)
+        .contains(ServiceBusKcqlProperties.SERVICE_BUS_TYPE);
```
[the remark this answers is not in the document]
GoMati-MU: Agreed, thanks!

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/ServiceBusReceiverFacadeTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+  @Test
+  void checkReceiverInitializationAndReceiverIdForQueue() {
+    //given
+    String queueType = "QUEUE";
+    Map<String, String> propertiesMap = mock(Map.class);
+    when(propertiesMap.get(ServiceBusKcqlProperties.SERVICE_BUS_TYPE.getPropertyName())).thenReturn(queueType);
+
+    Kcql kcql = mockKcql(propertiesMap);
+    String receiverId1 = UUID.randomUUID().toString();
```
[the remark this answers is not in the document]
GoMati-MU: Changed

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/ServiceBusReceiverFacadeTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+  @Test
+  void shouldGetIllegalArgumentExceptionOnCreationWithBadlyFormattedConnectionString() {
+    //given
+    Kcql kcql = mock(Kcql.class);
+    String badFormatConnectionString = "connectionString";
+
+    //when
+    assertThrows(IllegalArgumentException.class, () -> new ServiceBusReceiverFacade(kcql, mockedQueue,
+        badFormatConnectionString, UUID.randomUUID().toString())
```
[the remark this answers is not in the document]
GoMati-MU: Changed

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridgeTest.java
```diff
@@ -0,0 +1,140 @@
...
+    testObj.commitRecordInServiceBus(sourceRecord);
+
+    //then
+    verify(executorService).submit(any(Runnable.class));
+  }
+
+  private ServiceBusReceiverFacade mockReceiver() {
+    ServiceBusReceiverFacade receiverFacade = mock(ServiceBusReceiverFacade.class);
+
+    return receiverFacade;
```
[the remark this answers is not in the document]
GoMati-MU: Changed

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
GoMati-MU: My bad here, I was thinking you're simply intending to move `createRecievers` to the Task. I've added it with small modifications, thanks!

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnector.java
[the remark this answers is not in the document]
GoMati-MU: In this approach we’re creating `maxTasks` additional immutable objects where we only need one. I tweaked it a little not to go down that path. If it doesn't suit us, please let me know and I will revert to the original suggestion, cause this one gets way more attention than is should really ever get IMO.

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/util/KcqlConfigBusMapper.java
GoMati-MU: I splitted the mapping and validation logic to make it more clear (thanks for all advices here!)

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnector.java
GoMati-MU: Debate about Exceptions and whether to handle them or not is going on for years now, even Oracle having [their stance](https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html) and autority figures like Robert Martin having [different ones](https://santim0ren0.medium.com/clean-code-error-handling-2f973908b13e).  (_I tend to be more with Martin on it especially after introducing streams to Java, but it's not place nor the time_)

This way or another, this is a classic scenario of unrecoverable situation - Connector is having bad configuration passed to it, we want to fail fast. That's why I think there's no sense in using checked exception, nor create all the boilerplate (`try-catch` and `throws` for all the methods up to the Connector itself).

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnector.java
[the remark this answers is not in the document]
GoMati-MU: To be completely honest: this changes nothing in terms of duplication as before we had one line:
`new JarManifest(getClass().getProtectionDomain().getCodeSource().getLocation());`
and now we'll have one line:
`JarManifest.produceFromClass(getClass());`

But it's counterproductive to stay on it for more time, so I just added the util.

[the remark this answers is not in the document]
davidsloan: Seems like an improvement to me, thank you 👏

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/TaskToReceiverBridge.java
davidsloan: It's ok, happy to leave it as it is.  Thanks for answering my question 🥇

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnector.java
davidsloan: Absolutely valid sources.  Thank you for sharing.  But as said previously we wish to consider a more functional approach to our code that reduces side-effects and makes the flow clearer staying away from unintended consequences.  We've taken this discussion into the team channel and will update when we have a resolution. 👍

java-connectors/kafka-connect-azure-servicebus/src/test/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnectorTest.java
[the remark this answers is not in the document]
davidsloan: Using `Map.of()` in unit tests ensures the map is immutable, enhancing test consistency and safety by preventing unintended modifications. This approach also improves code clarity and conciseness, making it a beneficial change for more reliable tests.  Using immutable structures is considered best practice in a more functional style of programming, promoting safer and more predictable code.

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/util/KcqlConfigBusMapper.java
GoMati-MU: Done (i)

java-connectors/kafka-connect-azure-servicebus/src/main/java/io/lenses/streamreactor/connect/azure/servicebus/source/AzureServiceBusSourceConnector.java
davidsloan: I'm going to resolve this for now and will raise a technical debt ticket to address when we have decided functional error handling approach.

davidsloan:
Merged on dev repo.

PR #1234, opened 2024-05-22
fix to prefer tag before build script version for java JARs

GoMati-MU:
Fix to prefer latest git tag before build script version when building java connector JARs

### DIFF
--- a/java-connectors/build.gradle
+++ b/java-connectors/build.gradle
@@ -94,11 +94,10 @@ allprojects {
     }
 
     shadowJar {
-        def artifactVersion = (gitTag == null || gitTag.isBlank()) ?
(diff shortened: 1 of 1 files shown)


java-connectors/build.gradle
```diff
@@ -94,6 +94,8 @@ allprojects {
     }
 
     shadowJar {
+        def artifactVersion = (gitTag == null || gitTag.isBlank()) ?
+                project.version : gitTag
```
davidsloan, line 98: I think this would do what you want, but I haven't tried it.

```suggestion
        def artifactVersion = gitTag?.trim() ? gitTag : project.version
```

davidsloan (approved):

PR #1293, opened 2024-07-02
Add support for source connector to control the poll interval from s3

asafglassbox:
Added support for controlling the poll interval on the source connector. The default poll interval is 1 second. This update allows the interval to be changed using the parameter connect.s3.poll.interval.ms (in millisecond)


stheppi:
@asafglassbox 

Thank you for the PR.

I believe the delay should be applied only when there are no records to be returned. Otherwise, it would unnecessarily delay the delivery of available records. We recommend implementing a backoff strategy. Instead of using thread sleep, you could return an empty list when no records are available:

```
if there are no more records, set the task in "backoff mode". Track the currentTimestamp + backoff_interval, and for any subsequent poll method calls within this interval, just return empty records.
```

This approach prevents blocking the thread in the Connect execution thread pool. According to the documentation, the poll method "should block but return control to the caller regularly (by returning null)."

Issue #799, opened 2021-09-16
MQTT Source - Configurable Delimiter in Kafka Topic


Kiir080:
Hi,

I'm trying to use MQTT Source connector to mirror a set of topics of a MQTT Broker to Kafka.

In order to do that, I'm using the dynamic targets feature, as implemented in #650, and subscribing the MQTT topics using a wildcard with the following kcql statement: INSERT INTO \`$\` SELECT * FROM device/+/variable/+

Currently, the resulting kafka topic will be constructed by removing the first occurrence of a `/`, in the mqtt topic, and the further `/` will be replaced by `_` as implied by the following code in [MqttManager.scala](https://github.com/lensesio/stream-reactor/blob/15a5cbaf1fe005643d45cdcf2c46e5953acafb76/kafka-connect-mqtt/src/main/scala/com/datamountaineer/streamreactor/connect/mqtt/source/MqttManager.scala#L94):

```
  val kafkaTopic = kcql.getTarget match {
      case "$" => topic.replaceFirst("/", "").replaceAll("/+", "_").replaceAll("/", "_")
      case other => other
    }
```
In my particular example, I get something like this:
**MQTT Topic:** device/device1/variable/var1
**Obtained Kafka Topic:** devicedevice1_variable_var1

Would it be considerable to extend this functionality to customize the delimiter of the resulting Kafka topic and also remove the need to start the MQTT topics with `/` ?

PR #1673, opened 2025-05-09
fix(kafaka-connect-mqtt): Only remove leading slash on target

Izzette:
* Leading slash from MQTT topic should be removed when converting to Kafaka topic target.
* If no leading slash in MQTT topic exists, the first slash should _not_ be stripped and instead replaced with and underscore, as with the remaining slashes.

For example:

MQTT topic `/I/Love/Kafka` should be converted to `I_Love_Kafka`.
However, MQTT topic `Foo/Bar/Baz` should not be converted to `FooBar_Baz` but instead to `Foo_Bar_Baz`.

Relates to https://github.com/lensesio/stream-reactor/issues/799

I would like to add tests, but there seem to be no test cases around `$` targets, and I am a little unsure of how to instrument the connector for unit-testing.  Some guidance from maintainers would be appreciated.

### DIFF
--- a/kafka-connect-mqtt/src/main/scala/io/lenses/streamreactor/connect/mqtt/source/MqttManager.scala
+++ b/kafka-connect-mqtt/src/main/scala/io/lenses/streamreactor/connect/mqtt/source/MqttManager.scala
@@ -101,7 +101,7 @@ class MqttManager(
       )
 
     val kafkaTopic = kcql.getTarget match {
-      case "$"   => topic.replaceFirst("/", "").replaceAll("/+", "_").replaceAll("/", "_")
+      case "$"   => topic.replaceFirst("^/", "").replaceAll("/+", "_").replaceAll("/", "_")
       case other => other
     }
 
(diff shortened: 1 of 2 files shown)


stheppi:
Hi @Izzette 

Thank you for the PR. 

As an idea, the code you're changing

```
 val kafkaTopic = kcql.getTarget match {
      case "$"   => topic.replaceFirst("^/", "").replaceAll("/+", "_").replaceAll("/", "_")
      case other => other
    }
```

can be extracted as an util function which can be tested on it's own. And the tests can create a few KCQL statements like this:

```
Kcql.parseMultiple("INSERT INTO abc SELECT * FROM xyz; INSERT INTO def SELECT * FROM zyx;");
```
Once you have the KCQL instance you can pass it through the util function and assert on the output.

davidsloan (approved):

davidsloan:
Thank you for the contribution, I have approved it and we will merge it on our private repo to sync the changes.

Izzette:
@davidsloan / @stheppi thank you so much :)

Izzette:
It's been merged :)

Issue #1823, opened 2025-08-12
S3 Sink: broken in versions >= 8.1.33


sbuliarca:
Hi!

We're doing a backup of all topics using Parquet to an **generic S3 bucket**. Here are the details:

**Stream-reactor-version**: 8.1.30
**Kafka connect version**: based on Docker image confluentinc/cp-kafka-connect:7.8.0
**Backup config**:

{
"connector.class": "io.lenses.streamreactor.connect.aws.s3.sink.S3SinkConnector",
"tasks.max": "10",
"connect.s3.aws.auth.mode": "Default",
"connect.s3.compression.codec": "zstd",
"topics.regex": ".*",
"key.converter.schemas.enable": "false",
"connect.s3.kcql": "insert into msk-backup:msk-backup-parquet select * from * PARTITIONBY _topic, _partition STOREAS PARQUET PROPERTIES ('store.envelope'=true,'flush.count'=1000,'flush.interval'=300, 'flush.size'=50000000, 'partition.include.keys'=false);",
"name": "msk-s3-backup-sink-all-parquet",
"value.converter.schemas.enable": "false",
"connect.s3.compression.level": "9",
"connect.s3.aws.region": "eu-west-1",
"value.converter": "org.apache.kafka.connect.converters.ByteArrayConverter",
"key.converter": "org.apache.kafka.connect.converters.ByteArrayConverter"
}

**The Problem**:

We tried upgrading to 9.x and 10.x versions but the task failed with the following error:
```
ERROR Fatal error encountered while processing CopyOperation(msk-backup,.temp-upload/Topic(xxxxxx)/1/73a2b355-ce99-4cea-890c-2cf4a65a88d7msk-backup-parquet/xxxxxxx/1/xxxxxx(1_000000231500).parquet,msk-backup-parquet/xxxxxxx/xxxxxxxxx(1_000000231500).parquet,"523f5c485a7cbe944d58babae97907c3"): A header you provided implies functionality that is not implemented (Service: S3, Status Code: 501, Request ID: 530V91SMDRHZV0XD, Extended Request ID: 3rguAeLclKlBdMR4qWJVGKWLbzHg/QrSKZtv+DsFCQr0bcKN4nCiRxbSrD+wRPnoZMgPOLY2yfQ=) (io.lenses.streamreactor.connect.cloud.common.sink.seek.PendingOperationsProcessors)
```
[...]

PR #1851, opened 2025-08-27
Implement guard mechanism for sink tasks

dementiev27:
### Problem
As said in [cdc4e46](https://github.com/lensesio/stream-reactor/commit/cdc4e46a65306f2a26473284cc9577bb00a2d419), Kafka Connect can incorrectly run multiple tasks for the same topic/partition pair. Commit cdc4e46 introduced the ETag-based Upload-Copy-Delete flow to handle this. This works but **triples S3 API costs**.
For high-volume deployments, this makes the connector too expensive to run.

### Solution
Use Kafka transactions for distributed locking instead of the expensive upload flow.
Each task gets a unique transactional producer ID per topic/partition. Tasks send heartbeats to a compacted topic. If two tasks try to write to the same partition, Kafka's transaction coordinator automatically fences the older one.

### Config
The guard is disabled by default and can be enabled with:
```
# Enable guard functionality
connect.s3.guard.enable=true

# Kafka cluster for guard operations (required when enabled)
connect.s3.guard.bootstrap.servers=localhost:9092

# Guard heartbeat configuration (optional)
connect.s3.guard.heartbeat.topic=s3-sink-heartbeat
connect.s3.guard.heartbeat.ms=10000

# Topic auto-creation (optional)
connect.s3.guard.topic.autocreate.enable=true
connect.s3.guard.topic.autocreate.partitions=1
connect.s3.guard.topic.autocreate.replication.factor=3
```
Additional Kafka client properties can be configured using the `guard.client.*` prefix.

These changes allow removing the ETag-based Upload-Copy-Delete flow entirely, restoring direct uploads with full safety guarantees.

### DIFF
--- a/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTask.scala
+++ b/kafka-connect-aws-s3/src/main/scala/io/lenses/streamreactor/connect/aws/s3/sink/S3SinkTask.scala
@@ -15,18 +15,17 @@
  */
 package io.lenses.streamreactor.connect.aws.s3.sink
 
-import io.lenses.streamreactor.common.util.EitherUtils
-import io.lenses.streamreactor.common.util.JarManifest
+import io.lenses.streamreactor.common.util.{EitherUtils, JarManifest}
 import io.lenses.streamreactor.connect.aws.s3.auth.AwsS3ClientCreator
-import io.lenses.streamreactor.connect.aws.s3.config.S3ConfigSettings
-import io.lenses.streamreactor.connect.aws.s3.config.S3ConnectionConfig
+import io.lenses.streamreactor.connect.aws.s3.config.{S3ConfigSettings, S3ConnectionConfig}
 import io.lenses.streamreactor.connect.aws.s3.model.location.S3LocationValidator
 import io.lenses.streamreactor.connect.aws.s3.sink.config.S3SinkConfig
-import io.lenses.streamreactor.connect.aws.s3.storage.AwsS3StorageInterface
-import io.lenses.streamreactor.connect.aws.s3.storage.S3FileMetadata
+import io.lenses.streamreactor.connect.aws.s3.storage.{AwsS3StorageInterface, S3FileMetadata}
 import io.lenses.streamreactor.connect.cloud.common.config.ConnectorTaskId
+import io.lenses.streamreactor.connect.cloud.common.guard.KafkaGuardSupport
 import io.lenses.streamreactor.connect.cloud.common.sink.CloudSinkTask
-import io.lenses.streamreactor.connect.cloud.common.sink.WriterManagerCreator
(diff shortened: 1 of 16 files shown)


dementiev27:
Once the idea is approved, I'll cover changes with unit tests :)

dementiev27:
FYI – removing the Upload-Copy-Delete flow afterwards will also resolve #1823.

stheppi:
I don't think the solution is 100% guaranteeing the exactly once on S3. Ultimately that's the goal.

Here's the "zombie consumer" scenario and why I think the Kafka transactions don't help:

* Rebalance Starts: Task A is processing data for topic-partition-0. A network glitch causes a rebalance, and the coordinator revokes the partition from Task A and gives it to a new task, Task B.

* Task A Becomes a Zombie: Task A is slow to shut down and doesn't know it has lost the partition. It continues processing data.

* Transaction Fencing: Task A tries to commit its transaction. The Kafka broker sees that Task A's producer has been fenced (because a new producer from Task B registered with the same transactional.id). The broker rejects Task A's commit.

**The Flaw**: The transaction successfully prevented the zombie from writing bad data back into Kafka. However, the transaction has absolutely no ability to stop the zombie Task A's code from continuing its step, which is making an API call to write to S3. Task A might have been writing to S3 while Task B got ownership and processed overlapping data.

dementiev27:
@stheppi, in this implementation every file write is gated by a synchronous pre-write transactional heartbeat. If a newer task has taken over, the broker fences the old producer, the heartbeat fails, and we raise the exception to exit before making any S3 call. This gives a practical single-writer guarantee.
However, the only remaining gap is the tiny window after a successful heartbeat and before the PUT.

So you are right that it does not guarantee an exact 100% write-once on S3, but it is pretty close to it. In my opinion, many users would choose this option over significantly increasing API costs.

What do you think about making it configurable so users can choose between this transaction-based and an eTag-based solution?

stheppi:
> So you are right that it does not guarantee an exact 100% write-once on S3, but it is pretty close to it. In my opinion, many users would choose this option over significantly increasing API costs.

Thank you for your contribution and for engaging in this important discussion. You've raised a very valid point about the trade-offs between API costs and delivery guarantees, and we appreciate you highlighting it.

Our perspective is that when a feature is labeled "exactly-once," it must provide a strict guarantee. For many of our customers, especially in production environments, a "pretty close" guarantee can lead to significant data integrity issues. Misleading users with a best-effort implementation under an exactly-once flag is a risk we cannot take.

Therefore, we believe there are two clear paths:

* Enable full exactly-once semantics (the default), understanding the associated API costs.
* Set connect.s3.exactly.once.enable to false for at-least-once delivery and handle deduplication downstream if the cost of the first option is a concern.

We would be happy to discuss the requirements and potential further implementation to the at-least once that is improving the cost. If you're interested in exploring this further or discussing commercial support to accelerate its development, please feel free to reach out to us at sales@lenses.io.

Thanks again for your valuable input!

image

PR #1985, opened 2025-11-21
upgrade to antlr 4.13.1

vgw-chriskruger:
This PR will upgrade antlr to version 4.13.1. This should ostensibly address this issue raised here https://github.com/lensesio/fast-data-dev/issues/201


stheppi:
hi @vgw-chriskruger ,

thanks for the PR. we will look to exercise this on different kafka versions. we had users reporting issues around the ANTLR, so we need to understand the impact.

PR #2013, opened 2025-12-15
Fix connect.s3.vhost.bucket inverted semantics

klaudworks:
## Summary

- Fix inverted logic where `connect.s3.vhost.bucket` was passed directly to `pathStyleAccessEnabled()` which has opposite semantics
- `vhost.bucket=true` now correctly enables virtual-hosted style URLs (`bucket.endpoint/key`)
- `vhost.bucket=false` now correctly enables path-style URLs (`endpoint/bucket/key`)

## Problem

The `connect.s3.vhost.bucket` property in `AwsS3ClientCreator.scala` was being passed directly to `S3Configuration.builder().pathStyleAccessEnabled()`, but these have opposite meanings:

- `pathStyleAccessEnabled(true)` → Use **path-style** URLs: `http://endpoint/bucket/key`
- `pathStyleAccessEnabled(false)` → Use **virtual-hosted** URLs: `http://bucket.endpoint/key`

Meanwhile, the property name `vhost.bucket` implies:

- `vhost.bucket=true` → Enable **v**irtual **host** buckets
- `vhost.bucket=false` → Disable virtual host buckets (i.e., use path-style)

Because the code passed the value directly without inverting it, users with S3-compatible storage (MinIO, s3proxy, etc.) that only supports path-style access had to set `vhost.bucket=true` (counter-intuitive) to get it working.

## Fix

Invert the boolean before passing to `pathStyleAccessEnabled()`:
```scala
.pathStyleAccessEnabled(!config.enableVirtualHostBuckets)
```

## Breaking Change Note

Users who were relying on the buggy behavior will need to invert their `connect.s3.vhost.bucket` setting:
- If you had `connect.s3.vhost.bucket=true` for path-style access, change to `false`
- If you had `connect.s3.vhost.bucket=false` for virtual-hosted access, change to `true`

## Testing

Updated all integration tests that use S3 proxy containers to use `vhost.bucket=false` (path-style) which is required for S3-compatible emulators.

### DIFF
--- a/kafka-connect-azure-datalake/src/main/scala/io/lenses/streamreactor/connect/datalake/storage/DatalakeStorageInterface.scala
+++ b/kafka-connect-azure-datalake/src/main/scala/io/lenses/streamreactor/connect/datalake/storage/DatalakeStorageInterface.scala
@@ -343,7 +343,7 @@ class DatalakeStorageInterface(connectorTaskId: ConnectorTaskId, client: DataLak
             null,
             Context.NONE,
           )
-          (resp.getDeserializedHeaders.getETag, new String(baos.toByteArray))
+          (new String(baos.toByteArray), resp.getDeserializedHeaders.getETag)
       }
     }.toEither.leftMap {
       case ex: DataLakeStorageException if ex.getStatusCode == 404 =>
--- a/kafka-connect-azure-datalake/src/test/scala/io/lenses/streamreactor/connect/datalake/storage/DatalakeStorageInterfaceTest.scala
+++ b/kafka-connect-azure-datalake/src/test/scala/io/lenses/streamreactor/connect/datalake/storage/DatalakeStorageInterfaceTest.scala
@@ -532,7 +532,7 @@ class DatalakeStorageInterfaceTest
     }
     val result = storageInterface.getBlobAsStringAndEtag(bucket, path)
 
-    result.value should be((expectedEtag, expectedContent))
+    result.value should be((expectedContent, expectedEtag))
   }
 
   "writeStringToFile" should "upload the data string to the specified path when successful" in {


klaudworks:
## Alternative: Backwards-Compatible Fix

If the breaking change is a concern, I can implement an alternative backwards-compatible solution by introducing a new config parameter:

### Proposal

Add a new property `connect.s3.path.style.access` that:
1. Has clear, unambiguous semantics matching the AWS SDK's `pathStyleAccessEnabled()`
2. Takes priority over the existing `connect.s3.vhost.bucket` when set
3. Allows `connect.s3.vhost.bucket` to remain with its current buggy behavior for existing users

### Migration Path

1. New users should use `connect.s3.path.style.access=true` for S3-compatible storage
2. Existing users can continue using `connect.s3.vhost.bucket=true` (works as before)
3. Deprecation warning logged when `vhost.bucket` is used without `path.style.access`
4. Eventually deprecate `vhost.bucket` in a future major version

Let me know if you'd prefer this approach and I'll update the PR accordingly.

stheppi:
Hi @klaudworks ,
Thank you for the PR. I think compatibility is a concern, so your suggestion about the extra field sounds like a good solution. We would welcome your updated PR. 

Thanks

klaudworks:
@stheppi I just updated the PR to implement a new config parameter. This keeps backwards compatibility with the old buggy behavior.
After merging this someone should update the docs [docs.lenses.io/latest/connectors/kafka-connectors/sources/aws-s3](https://docs.lenses.io/latest/connectors/kafka-connectors/sources/aws-s3).

stheppi:
@klaudworks 
Much appreciated. We will release probably next week.

stheppi:
merged via the internal repo.

Thank you for your contribution.

PR #2017, opened 2025-12-18
provide a custom file namer

mkeskells:
Minimal change to provide support for custom file naming

This allows for a custom `FileNamerFactory` to be created (with an optional configuration parameter) 
The `FileNamerFactory` has a single method to create a  `FileNamer` (based on the parameters that the system already has for file namers

A `FileNamer` is an existing interface, and was minimally changed to pass parameters via a case class to limit compatibility issues should additional parameters be supported in the future

TODO

- [ ] Discuss this approach and refine with the maintainers
- [ ] update javadoc to match the modified APIs
- [ ] add tese to the configuration extraction
- [ ] add tests to the custom file buidler

To Explore
It would be good to allow more behaviour to the file naming, for instance 
- to include the start offset, number of records etc
- to ensure hat we never overwrite a file, add a suffix (this would be a change to the behaviour of the `FileNamer` I believe


stheppi:
merged via the internal repo.

Thank you for your contribution.

PR #2038, opened 2026-01-10
Upload file to final location. 

mkeskells:
Avoid copy and delete operation to reduce latency
Avoid the possibility of copying the incorrect file due to collision of the `tempFileName `

### DIFF
--- a/kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/writer/Writer.scala
+++ b/kafka-connect-cloud-common/src/main/scala/io/lenses/streamreactor/connect/cloud/common/sink/writer/Writer.scala
@@ -138,9 +138,21 @@ class Writer[SM <: FileMetadata](
         for {
           key         <- objectKeyBuilder.build(uncommittedOffset, earliestRecordTimestamp, latestRecordTimestamp)
           path        <- key.path.toRight(NonFatalCloudSinkError("No path exists within cloud location"))
-          pendingOperations = NonEmptyList.of[FileOperation](
-            UploadOperation(key.bucket, file, path),
-          )
+          pendingOperations = if (indexManager.indexingEnabled) {
+            val tempFileUuid = UUID.randomUUID().toString
+            val tempFileName = path.prependedAll(
+              s".temp-upload/${topicPartition.topic}/${topicPartition.partition}/$tempFileUuid",
+            )
+
+            NonEmptyList.of[FileOperation](
+              UploadOperation(key.bucket, file, tempFileName),
+              CopyOperation(key.bucket, tempFileName, path, "placeholder"),
+              DeleteOperation(key.bucket, tempFileName, "placeholder"),
+            )
+          } else
+            NonEmptyList.of[FileOperation](
+              UploadOperation(key.bucket, file, path)
+            )
 
           _ <- pendingOperationsProcessors.processPendingOperations(
             topicPartition,


stheppi:
I will start with the conclusion and explain why after.


## Conclusion

**It is NOT possible to safely change to single direct upload** while maintaining exactly-once semantics.

### Why the 3-Operation Approach Works

1. **UUID prevents destination conflicts**: Each attempt uses a unique temp path
2. **Copy can overwrite**: Enables recovery when file exists at destination
3. **eTag chain ensures integrity**: Only the intended temp file gets committed
4. **Atomic commit**: The final index update with `committedOffset = pendingOffset` is the commit point

### The Temp File Serves as a "Transaction Log"

- Upload to temp = "prepare"
- Copy to destination = "commit"  
- Delete temp = "cleanup"

The eTag links these operations, ensuring only complete, verified data reaches the destination.



## Why it is not possible 

### Scenario: Task Crash After Upload, Before Index Commit

```mermaid
sequenceDiagram
    participant TaskA as Task A (original)
    participant TaskB as Task B (after rebalance)
    participant Index as Index File
    participant Cloud as Cloud Storage
    
    Note over TaskA: Processing offsets 100-200
    
    rect rgb(255, 230, 230)
        Note right of TaskA: DIRECT UPLOAD APPROACH
        TaskA->>Index: PendingState: [Upload(dest)]
        TaskA->>Cloud: Upload to destination path
        Note over Cloud: File exists at destination
        TaskA->>TaskA: CRASH before index update
    end
    
    Note over TaskB: Rebalance - Task B takes partition
    TaskB->>Index: Read index
    Note over Index: committedOffset=100<br/>pending=[Upload(dest)]
    
    TaskB->>TaskB: Local file doesn't exist
    Note over TaskB: NonExistingFileError<br/>cancelPending=true
    
    TaskB->>Index: Cancel pending, keep committedOffset=100
    TaskB->>TaskB: Reprocess offsets 100-200
    TaskB->>TaskB: Generate same destination path
    TaskB->>Cloud: Upload to destination path
    
    Note over Cloud: FAILS! File exists<br/>ifNoneMatch("*") rejects
    
    Note over TaskB: UNRECOVERABLE STATE
```



### Why Direct Upload Breaks Exactly-Once

| Failure Mode | Current (3-op) | Direct Upload |
|--------------|----------------|---------------|
| Upload succeeds, crash before index update | New worker retries with NEW temp UUID, copy overwrites destination | New worker retries to SAME destination, `ifNoneMatch` fails |
| Concurrent workers during rebalance | Different temp UUIDs prevent conflict | Same destination path causes race |
| Recovery after partial commit | Copy can overwrite destination | Cannot retry upload to existing file |

### Root Cause

The combination of:

1. **Deterministic destination paths** (based on offset/timestamp)
2. **`ifNoneMatch("*")` preventing overwrites**
3. **Local file not available after rebalance**

Creates an **unrecoverable state** where the data exists at destination but cannot be acknowledged.

### Could We Remove `ifNoneMatch("*")`?

Removing `ifNoneMatch` would fix recovery but introduces a **race condition**:

```mermaid
sequenceDiagram
    participant TaskA as Task A
    participant TaskB as Task B
    participant Cloud as Cloud Storage
    
    Note over TaskA,TaskB: Both assigned same partition during rebalance transition
    
    TaskA->>Cloud: Upload data for offset 200 (v1)
    TaskB->>Cloud: Upload data for offset 200 (v2) - OVERWRITES
    TaskA->>TaskA: Commits offset 200
    Note over Cloud: File contains v2 data<br/>but TaskA thinks v1 committed
    
    Note over Cloud: DATA INCONSISTENCY
```

stheppi:
@mkeskells 

we will do the change for non-exactly once and one operation.

mkeskells:
Hi @stheppi 
Thanks for the deatiled explaination. I had though there there must be a reason to for 3 stage process, but missed it

I have updatd the PR to take account of the exactly once semantics (as I understand them)

mkeskells:
Hi @stheppi 
It looks like you have cherry picked thus now and merged to main

If so can you confirm and close this PR

PR #2039, opened 2026-01-10
upload files in parallel

mkeskells:
Allow files upload to occur on background threads

TODO - the parallelism and timeout isnt configurable, and for certain the current code in this PR, linked to the Runtime.availableCpus isnt viable, but I am unsure where the parallelism limits (if any) should be configured

- The code in this PR _could_ be driven by the environment variable for scala concurrency, but this is against the spirit of the way that Kafka Connect  in configured
- We could have no limit on parallelism, but this seems dangerous
- We could limit to an arbetry high value (e.g. 1000) 
- We could have a limit at the VM (clasloader) or task level

Too many questions to answer, so I defer to the maintainers for some input


stheppi:
Thanks for raising this. While the feature, the change leads to concurrent calls to IndexManager.update(...) are a real problem since the IndexManager implementation is not thread-safe and uses shared mutable state without synchronization.

We can look into making that change

mkeskells:
HI @stheppi 
I had a quick look at the indexmanager, and updated the `seekedOffset` and `topicPartitionToETags` to be a `TrieMap`
I am not sure if you had other concerns for the thread safety of the access, but his is I hope at least part of the change

mkeskells:
I havent address the configuration of the parallelism though

mkeskells:
I have not addressed the configuration of the parallelism

PR #2040, opened 2026-01-11
more flexible file naming

mkeskells:
Include name based on firstOffset and recordCount


stheppi:
Hi @mkeskells 

Please help me understand the value of having the first offset written. Is it for analytics on your end? My understanding is that the latest release fixes the data loss.

Also having first and last record written makes the record count redundant no? it can be computed based on the two offsets parameters.

mkeskells:
HI @stheppi 
Depending on the configuration not all records between offset start and offset end may be written to the same file. 
If for instance the records are stored in a HIVE partitiion keyed by a value in the headers/payload, or there are different schema

Also it was just to allow the flexibility to enable the file namer to choose whatever fields are appropriate fo the naming scheme a user may want.

I thought it would be useful to allow some flexibility. Ideally it would be good to allow some configuration to supply MetaData to be attached to the blob, which can be used by some downstream process. The filename is a simplified version of that. It allows information to be retrieved about the content, without reading the blob content

Issue #1386, opened 2024-10-01
MQTT source - possibility to add prefix to kafka topic


dberardo-com:
i want to have 2 source connectors reading from the same mqtt topic and writing to 2 different kafka topics. why i need this? because i am using a different partitioning key

is it possible to a add a prefix so "select * from" to add to the kafka topic name ?

stheppi:
HI @dberardo-com 

Not at the moment. you could use an SMT to adapt the topic name.

dberardo-com:
Thanks for the reply.

So are you suggesting to create 2 connectors, 1 of which would use an SMT to rewrite the topic name? 

As I am using wildcard mqtt subscriptions, di you know any SMT that can rewrite topic names dynamically (e.g. using regex or so)? Or maybe simply appending a suffix "_copy" to the topic names?

Woestebanaan:
I got a fix for this: https://github.com/lensesio/stream-reactor/pull/2149

PR #2149, opened 2026-04-19
MQTT wildcard and prefix fix

Woestebanaan:
<p><strongMQTT wildcard and prefix fix</strong></p>
<p>Previously, the <code>$</code> wildcard in KCQL target mappings only worked as a standalone value — <code>getTarget</code> had to return exactly <code>"$"</code> for the MQTT source topic to be used as the target. If a user wrote something like <code>"prefix_$"</code> or <code>"data/$_events"</code>, the <code>$</code> would not be substituted and the literal string (including <code>$</code>) would be used as-is.</p>
<p>This change adds support for <code>$</code> as an embedded placeholder within a larger target string. The sanitized MQTT topic name is now substituted wherever <code>$</code> appears in the target.</p>
<p><strong>Examples:</strong></p>

KCQL Target | MQTT Topic | Before | After
-- | -- | -- | --
$ | /sensor/temp | sensor_temp | sensor_temp
prefix_$ | /sensor/temp | prefix_$ | prefix_sensor_temp
data/$_events | /sensor/temp | data/$_events | data/sensor_temp_events

Also fixed the single and multi-level MQTT wildcards

Resolved the following issue: [#1386 ](https://github.com/lensesio/stream-reactor/issues/1386)

### DIFF
--- a/kafka-connect-mqtt/src/main/scala/io/lenses/streamreactor/connect/mqtt/source/MqttManager.scala
+++ b/kafka-connect-mqtt/src/main/scala/io/lenses/streamreactor/connect/mqtt/source/MqttManager.scala
@@ -69,7 +69,7 @@ class MqttManager(
     actualTopic.matches(
       subscribedTopic.replaceAll("\\$share/.*?/", "")
         .replaceAll("\\+", "[^/]+")
-        .replaceAll("#", ".+")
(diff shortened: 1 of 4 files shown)


stheppi:
Merged. @Woestebanaan thanks for contributing

PR #2162, opened 2026-05-05
fix(cloud-sink): widen int8/int16 values for Avro int

klaudworks:
## Problem
Kafka Connect INT8 and INT16 schemas are mapped to Avro int, but the runtime values backing those schemas can be java.lang.Byte or java.lang.Short. Those values are not widened to java.lang.Integer before writing the Avro record, so Avro rejects them for int fields with errors like:

`AvroRuntimeException: Unknown datum type java.lang.Byte`

## Fix
When the target Avro schema type is INT, widen Byte and Short values to Int before writing the GenericRecord.

## Testing
Added regression coverage for nullable Avro int fields backed by Connect INT8 and INT16 schemas.

Also validated the fix in an S3 sink deployment that was failing while serializing java.lang.Byte values.


stheppi:
Merged. @klaudworks thanks for your contribution.

